PR #12, opened 2020-08-28
Update operand version to v0.6.0

ArangoGutierrez:
Also Add  missing nodeSelector for worker daemonSet 

### DIFF
--- a/manifests/0600_operator.yaml
+++ b/manifests/0600_operator.yaml
@@ -36,7 +36,7 @@ spec:
             - name: OPERATOR_NAME
               value: "node-feature-discovery-operator"
             - name: NODE_FEATURE_DISCOVERY_IMAGE
-              value: "quay.io/kubernetes_incubator/node-feature-discovery:v0.5.0"
+              value: "k8s.gcr.io/nfd/node-feature-discovery:v0.6.0"
           securityContext:
             readOnlyRootFilesystem: true
             allowPrivilegeEscalation: false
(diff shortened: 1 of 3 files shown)


manifests/0600_operator.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             - name: OPERATOR_NAME
               value: "node-feature-discovery-operator"
             - name: NODE_FEATURE_DISCOVERY_IMAGE
-              value: "quay.io/kubernetes_incubator/node-feature-discovery:v0.5.0"
+              value: "quay.io/kubernetes_incubator/node-feature-discovery:v0.6.0"
```
marquiz, line 39: I think we could/should use `k8s.gcr.io/nfd/node-feature-discovery:v0.6.0`

marquiz (commented):
Thanks! Could you squash these into one commit? Please also capitalize the first word of the message subject line

Good read 😄 
https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines

ArangoGutierrez:
ups, Thanks @marquiz , will do

marquiz (commented):
This probably sounds like useless nitting but seems like the commit message just hangs in the air unfinished 😄  "Use xxx instead..." of what(?) Prolly "use k8s.gcr.io instead of quay.io"? Also, you could comment on the nodeSelector why it is added in this commit.

marquiz:
Thanks for your patience @ArangoGutierrez, looks good to me 😄 

/lgtm

PR #15, opened 2020-09-14
scripts: make push-image.sh runnable locally

marquiz:
Override VERSION makefile variable only if _GIT_TAG is specified (most
likely in prow).


ArangoGutierrez (approved):
/lgtm

PR #16, opened 2020-09-14
Use official golang image as BUILDER

ArangoGutierrez:
prow job
testgrid.k8s.io/sig-node-node-feature-discovery#operator-push-image is
failing, this might be because we are using golang from UBI which comes
with a default user that sometimes depending on the build environment
doesn't have write persmisions.


ArangoGutierrez:
also opening this PR made me realize we should bump the go version for this repo. will do after we can get PROW happy

marquiz (approved):
Thanks @ArangoGutierrez, works for me!
/lgtm

PR #17, opened 2020-09-14
Bumb golang version to 1.14.7

ArangoGutierrez:
NFD operand is using 1.14.7 already, the operator should be kept in sync


marquiz (approved):
Good!
/lgtm

PR #25, opened 2020-09-17
Add RELEASE.md

marquiz:
An adaptation of RELEASE.md from node-feature-discovery.


ArangoGutierrez (approved):
we are so close!!! excited, thanks to everyone involved
@marquiz should we target for a release cadence? every 3 months or so?
/lgtm

PR #27, opened 2020-09-21
Update default operand image

ArangoGutierrez:
Walking towards a first release of the Operator, we should sync Operand latest release with the latest (not yet released) version of the operator 
moving from temporal registry quay.io/kubernetes_incubator/node-feature-discovery:latest -> k8s.gcr.io/nfd/node-feature-discovery:v0.6.0

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


marquiz:
Looks good. Could you remind me where/when is this value actually used? We deploy the operand using the templates from `assets/`, right(?)
/lgtm

Issue #30, opened 2020-09-23
make clean-labels fails 


ArangoGutierrez:
**What happened**:

```bash
eduardo@fedora-32-ws node-feature-discovery-operator]$ make clean-labels                                               
make: node/ip-172-20-111-4.ec2.internal: Command not found
make: *** [Makefile:77: clean-labels] Error 127
```

**What you expected to happen**:

```bash
[eduardo@fedora-32-ws node-feature-discovery-operator]$ make clean-labels                                               
kubectl get no -o yaml | sed -e '/^\s*nfd.node.kubernetes.io/d' -e '/^\s*feature.node.kubernetes.io/d' | kubectl replace -f -
node/ip-172-20-111-4.ec2.internal replaced
node/ip-172-20-34-84.ec2.internal replaced
node/ip-172-20-49-125.ec2.internal replaced
node/ip-172-20-91-201.ec2.internal replaced
node/ip-172-20-92-84.ec2.internal replaced
node/ip-172-20-96-110.ec2.internal replaced
```

**How to reproduce it (as minimally and precisely as possible)**:
attempt to run `make clean-labels`

Issue #29, opened 2020-09-23
Enhance nfd-worker placement


ArangoGutierrez:
**What would you like to be added**:
enable NFD operator to deploy nfd-worker pods on nodes labeled other than "node-role.kubernetes.io/worker"

**Why is this needed**:

Current implementation of the NFD-operator create a daemonset which is only taking nodes with "node-role.kubernetes.io/worker" into accout. Kubernetes deployments without this label on non-master nodes won't see nfd-worker scheduled.

marquiz:
I left this same comment in #31: Do we really need some affinity setting? Daemonset should be scheduled on on every schedulable node, shoudln't it(?)

Issue #19, opened 2020-09-17
Operand configurability


marquiz:
A meta-issue (inspired by #18) where I tried to collect things that should be configurable. The mechanism to use would be the operand CRD.

Master:
- [ ] mTLS configuration 
  - [ ] `--ca-file`, `--cert-file` and `--key-file`
  - [ ] `--verify-node-name`
- [ ] `--extra-label-ns`
- [ ] `--resource-labels`
- [ ] support `--prune` in some way to help with un-installation??

Worker:
- [ ] mTLS configuration via `--ca-file`, `--cert-file` and `--key-file`
- [x] user-specified worker config through ConfigMap (#18, #36)
- [x] `--sleep-interval`

mythi:
I think many of these could be covered by supporting configurable args for both master and worker:

```
type NodeFeatureDiscoverySpec struct {
	OperandNamespace    string `json:"operandNamespace"`
	OperandImage             string `json:"operandImage"`
        OperandMasterArgs     []string `json:"operandMasterArgs"`
        OperandWorkerArgs     []string `json:"operandWorkerArgs"`

}
```

Perhaps TLS should be managed by cert-manager?

marquiz:
> I think many of these could be covered by supporting configurable args for both master and worker:

Sure, they're just hard to verify and so easy to mess up. I don't know if this is a usual pattern in the "operator world", though. I don't think it would be much harder to add separate fields for the important command line flags. Maybe the args could be a first aid, though and useful in debugging or very exotic use cases(?)



> Perhaps TLS should be managed by cert-manager?

Yes it *should*, there's even an issue about that but I haven't had bandwidth for that.

ArangoGutierrez:
sorry for that, testing some prow things for this repo

ArangoGutierrez:
#36 will address `user-specified worker config through ConfigMap `

Issue #4, opened 2020-04-20
Run nfd-master as Deployment


marquiz:
Run nfd-master as Deployment instead of a DaemonSet, similar what the [NFD upstream deployment template](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/nfd-master.yaml.template) does.

Makes it possible to deploy NFD in managed clusters which have no master nodes.

ArangoGutierrez:
My concern with this concept is that worker nodes could tag them selves as Masters and gain privileges on the system. How could we avoid that on a Vanilla Kubernetes deployment?

marquiz:
How would a Deployment be different in terms of privileges??

ArangoGutierrez:
From the issue init comment "clusters which have no master nodes." , then a Deployment will create masters and a way to create labels without controlling the head of the cluster. I may be wrong tho, not a security expert by no means, I keep my password on a post-it. but I think is something worth double checking before going into that direction

marquiz:
Master nodes and nfd-master are a different/separate thing. When you run nfd-master it doesn't label the node it's running on as a master node. You need to run nfd-master somewhere in the cluster (in order for NFD to be able to work i.e. create node labels). If there are no master nodes in the cluster you'll end up with a non-working state when using a DaemonSet.

Nevertheless, the thinking behind running nfd-master on master node(s) by default was still security. Nfd-master would be "safer" there because master nodes are unlinkely to run any user workloads. In any case, I think the operator should take the "masterless" cluster into consideration, only use Deployment in that case or smth.

marquiz:
Still valid
/remove-lifecycle stale

ArangoGutierrez:
@zvonkok I would like to get your eyes on this issue

PR #31, opened 2020-09-23
Enhance nfd-worker placement

ArangoGutierrez:
Enable NFD operator to deploy nfd-worker pods on nodes labeled other
than "node-role.kubernetes.io/worker"

Also:  Fix clean-labels make target

Closes: #29
Closes: #30

### DIFF
--- a/assets/worker/0700_worker_daemonset.yaml
+++ b/assets/worker/0700_worker_daemonset.yaml
@@ -13,6 +13,10 @@ spec:
       labels:
         app: nfd-worker
     spec:
+      tolerations:
+      - operator: "Exists"
+        effect: "NoSchedule"
+
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
--- a/assets/worker/0700_worker_daemonset.yaml
+++ b/assets/worker/0700_worker_daemonset.yaml
@@ -16,7 +16,6 @@ spec:
       tolerations:
       - operator: "Exists"
         effect: "NoSchedule"
-
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
--- a/assets/worker/0700_worker_daemonset.yaml
+++ b/assets/worker/0700_worker_daemonset.yaml
@@ -13,8 +13,13 @@ spec:
       labels:
         app: nfd-worker
     spec:
-      nodeSelector:
-        node-role.kubernetes.io/worker: ""
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key:  node-role.kubernetes.io/master
+                operator: DoesNotExist
       hostNetwork: true
       serviceAccount: nfd-worker
       readOnlyRootFilesystem: true
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ clean:
 	rm -f $(BIN)
 
 clean-labels:
-	@$(shell kubectl get no -o yaml | sed -e '/^\s*nfd.node.kubernetes.io/d' -e '/^\s*feature.node.kubernetes.io/d' | kubectl replace -f -)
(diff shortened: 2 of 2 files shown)


marquiz (commented):
Hmm, I guess I never totally understood the old nodeSelector 🤔  Do we really need some affinity? By default (in upstream Kubernetes at least) daemonset should be scheduled on on every schedulable node, right(?)

The Makefile fix makes sense in any case.

ArangoGutierrez:
> Hmm, I guess I never totally understood the old nodeSelector thinking Do we really need some affinity? By default (in upstream Kubernetes at least) daemonset should be scheduled on on every schedulable node, right(?)
> 
> The Makefile fix makes sense in any case.

this PR affect only `nfd-worker` to be scheduled on non master nodes. by utilizing https://v1-16.docs.kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity.

I know we have #4 but that is for `nfd-master` (fix will come after this PR) , but this PR really target use cases where kubernetes "worker" nodes are being labelled something different than worker for different use cases, I have seen some users going for `infra` other for `test ` and so on, but the current nodeSelector is failing to schedule `nfd-worker` on those.

By Using nodeAffinity we can better help jube-scheduler to deploy `nfd-worker` pods

marquiz:
Mm, I know how node affinity works. But again, why do we need node affinity for workers? Shouldn't we just let it run on all (schedulable) nodes?
https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#running-pods-on-select-nodes

In some setups even master node might be configured to run regular workloads in which case this new affinity will prevent nfd-worker from running there.

zvonkok:
@marquiz There are several customers that apply taints to workers, since we do not know which taints, the NFD worker daemonset tolerates all taints, which means they can be also scheduled on masters. 

For now we want to prevent NFD to be scheduled on masters but I have seen requests for labelling masters as well. 

For now, tolerate all taints, and set an anti-affinity on the master. Let NFD worker be scheduled only on "any" worker.

marquiz:
> For now, tolerate all taints, and set an anti-affinity on the master. Let NFD worker be scheduled only on "any" worker.

OK, that explains. This patch would make sense IF there were some tolerations added. But, currently there are none.



> For now we want to prevent NFD to be scheduled on masters but I have seen requests for labelling masters as well.

Yet another configurability option for #19?

assets/worker/0700_worker_daemonset.yaml
```diff
@@ -13,8 +13,13 @@ spec:
       labels:
         app: nfd-worker
     spec:
```
marquiz, line 15:
```suggestion
    spec:
      tolerations:
      - operator: "Exists"
        effect: "NoSchedule"

```

ArangoGutierrez: Tested this patch and it makes the operator to run in a non desired state.

ArangoGutierrez: to be clear, you propose adding these lines without removing the other additions ?

ArangoGutierrez: If I add those lines,  both `nfd-master` and `nfd-worker` pods start shutting on and off 
```bash
NAME                            READY   STATUS        RESTARTS   AGE
nfd-master-22qbg                1/1     Terminating   0          23s
nfd-master-pfcgq                1/1     Running       0          2m39s
nfd-master-vt66q                1/1     Running       0          2m39s
nfd-operator-5d9955799d-npmx7   1/1     Running       0          2m41s
nfd-worker-dnxg7                1/1     Running       0          2m39s
nfd-worker-dwfpj                1/1     Running       0          2m39s
nfd-worker-tphv4                1/1     Terminating   0          81s
```

they go from Running to Terminating and back in a loop

marquiz: You have some other problem with your setup. Running two operators or smth? How would worker spec affect master??

And yes, add both tolerations and affinity. Affinity does not make any sense without the toleration. Agree?

marquiz (approved):
Thanks @ArangoGutierrez ! Now this makes sense 😉

Issue #18, opened 2020-09-17
immutable nfd-worker config


mythi:
the reconciler resets my changes to nfd-worker `ConfigMap`. How do I deploy the operator from operatorhub but add my own nfd-worker.conf?

ArangoGutierrez:
Heya @mythi could you test #36 and comment if it fixes this issue, thanks!

ArangoGutierrez:
https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/39

PR #36, opened 2020-11-10
Allow nfd-worker.conf to be modify after deployment 

ArangoGutierrez:
Current version of the NFD operator has a static config-map which configures the NFD-worker, and the controller is reconciling to the initial ConfigMap.

This patch modifies the handler to perform an Update function when the resource is modified 

Fixes:  #18

### DIFF
--- a/pkg/controller/nodefeaturediscovery/nodefeaturediscovery_controller.go
+++ b/pkg/controller/nodefeaturediscovery/nodefeaturediscovery_controller.go
@@ -60,6 +60,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to ConfigMap nfd-worker.conf
+	if err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
+
 	// Watch for changes to secondary resource Pods and requeue the owner NodeFeatureDiscovery
 	cache := &handler.EnqueueRequestForOwner{
 		IsController: true,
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
tested with  `quay.io/eduardoarango/node-feature-discovery-operator:8ea3ef2-dirty` 

```bash
IMAGE_TAG=quay.io/eduardoarango/node-feature-discovery-operator:8ea3ef2-dirty make deploy
# Edit the ConfigMap
kubectl edit cm/nfd-worker -n node-feature-discovery-operator
# Wait to the reconciler to kick off
sleep 30
# Check modifications are still in place
kubectl get cm/nfd-worker -n node-feature-discovery-operator -o yaml
```

pkg/controller/nodefeaturediscovery/nodefeaturediscovery_controller.go
```diff
@@ -60,6 +60,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to ConfigMap nfd-worker.conf
+	if err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return err
+	}
```
marquiz, line 66: Prolly a stupid question, but how does this change prevent the controller from overwriting the ConfigMap? Isn't the `Reconcile()` run whenever any of the watched resources are modified(?)

zvonkok: It is a good question, it will not prevent it. If you e.g. delete the DaemonSet the operator will overwrite and update the CM since a watched resource changed the the reconcile loop kicks in.

ArangoGutierrez:
I will try this again with a diff approach, closing this PR. 
Will create a new one with similar name

PR #37, opened 2020-11-13
master:worker nodes enablement for nfd-worker

ArangoGutierrez:
After https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/31 three node cluster, where all nodes will have the master and worker(or node for vanilla clusters) labels are just getting the nfd-master daemonset deployed. Since they have the master label, the NFD workers will not be scheduled.

This Patch fix that adding support for master:worker type of nodes, being used in edge deployments 

### DIFF
--- a/assets/worker/0700_worker_daemonset.yaml
+++ b/assets/worker/0700_worker_daemonset.yaml
@@ -14,15 +14,18 @@ spec:
         app: nfd-worker
     spec:
       tolerations:
-      - operator: "Exists"
-        effect: "NoSchedule"
+        - operator: "Exists"
+          effect: "NoSchedule"
       affinity:
         nodeAffinity:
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
@rmkraus FYI

ArangoGutierrez:
Discussion started in a downstream distribution of NFD: https://github.com/openshift/cluster-nfd-operator/issues/109

marquiz (commented):
Thanks @ArangoGutierrez ! The change ooks good.

My only gripe is about the commit message which is very minimal and confusing. E.g. in the description of this PR there is somewhat good background information which could be used (almost as is) in the commit message

marquiz (approved):
/lgtm

PR #39, opened 2020-11-19
Expand nodeFeatureDiscovery CRD

ArangoGutierrez:
This Patch expands the nodeFeatureDiscovery CRD to enable users to
define the Operand image provenance regirsty/image:version as well as
the imaePullPolicy, that by default has been set to "Always"

Also now the nodeFeatureDiscovery CRD holds the binaryData for the
NFD-Worker configuration file, allowing users to define the desired
nfd-worker.conf by editing the CR.

To achieve this we move the api from v1alpha1 to v1 and extend the CRD
and the controller to watch for the updated resources.

extra: go mod tidy removed 2 non needed deps from go.mod

Fixes: #18


ArangoGutierrez:
Fixes: #18

manifests/0700_cr.yaml
```diff
@@ -1,8 +1,26 @@
...
   name: nfd-master-server
   namespace: REPLACE_NAMESPACE
 spec:
-  namespace: node-feature-discovery-operator
-  image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+  operand:
+    namespace: node-feature-discovery-operator
+    image: node-feature-discovery
+    repository: k8s.gcr.io/nfd
+    version: v0.6.0
```
mythi, line 11: is there a use-case to keep these separate? My preference would be to keep just `image` . FWIW, I'm also not sure about `namespace` if the deployment gets created in the CR namespace.

manifests/0700_cr.yaml
```diff
@@ -1,8 +1,26 @@
...
+    imagePullPolicy: Always
+  nfd-worker:
```
marquiz, line 13: Should we use camelCase here, too, for consistencty, or not(?)

pkg/apis/nfd/v1/nodefeaturediscovery_types.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NodeFeatureDiscoverySpec defines the desired state of NodeFeatureDiscovery
+// +k8s:openapi-gen=true
+type NodeFeatureDiscoverySpec struct {
+	Operand OperandSpec `json:"operand"`
+	Config  ConfigSpec  `json:"nfd-worker"`
```
marquiz, line 12: If this is just for the worker config, should we name it `WorkerConfig` (and use`json:"workerConfig"`, respectively)?

pkg/apis/nfd/v1/nodefeaturediscovery_types.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Image pull policy
+	// +kubebuilder:validation:Optional
+	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
+}
+
+// ConfigSpec describes configuration options for the NFD worker
+type ConfigSpec struct {
+	// BinaryData holds the NFD configuration file
+	BinaryData string `json:"config"`
```
marquiz, line 37: I'd suggest to align the naming, e.g.
```
configData string `json:"configData"`
```
or something else with the same idea...

pkg/apis/nfd/v1/nodefeaturediscovery_types.go
```diff
@@ -0,0 +1,95 @@
...
+	case "Always":
+		imagePullPolicy = corev1.PullAlways
+	case "Never":
+		imagePullPolicy = corev1.PullNever
+	case "IfNotPresent":
+		imagePullPolicy = corev1.PullIfNotPresent
+	default:
+		imagePullPolicy = corev1.PullIfNotPresent
+	}
+	return imagePullPolicy
```
marquiz, line 89: Suggesting a simplification:

```suggestion
	switch corev1.PullPolicy(pullPolicy) {
	case corev1.PullAlways:
		return corev1.PullAlways
	case corev1.PullNever:
		return corev1.PullNever
	}
	return corev1.PullIfNotPresent
```

pkg/controller/nodefeaturediscovery/nodefeaturediscovery_controls.go
```diff
@@ -259,6 +257,10 @@ func ConfigMap(n NFD) (ResourceStatus, error) {
 
 	obj.SetNamespace(n.ins.GetNamespace())
 
+	// Update ConfigMap
+	obj.ObjectMeta.Name = "nfd-worker"
```
marquiz, line 261: Should we name this nfd-worker-conf?

marquiz (commented):
Thanks @ArangoGutierrez for the patch!  From what I can tell, this looks basically good, just few minor comments/suggestions.

BTW, try running `go mod tidy` again. For me it tidies go.mod and go.sum even more

manifests/0700_cr.yaml
ArangoGutierrez: This granularity help us during testing, is almost the same I know, but I prefer to visualize it this way. I am ok with a single var holding the image info.

The namespace is there as a placeholder, I am not modifying this, current CR has image and namespace

ArangoGutierrez:
Will squash once we agree on the patch state

manifests/0700_cr.yaml
```diff
@@ -1,8 +1,100 @@
...
-  namespace: node-feature-discovery-operator
-  image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+  operand:
+    namespace: node-feature-discovery-operator
+    image: node-feature-discovery
+    repository: k8s.gcr.io/nfd
+    version: v0.6.0
+    imagePullPolicy: Always
+  workerConfig:
+    config: |
```
marquiz, line 14: Should this be `configData` to reflect the latest changes(?)

ArangoGutierrez: epic face palm, it does

marquiz:
I only had one more comment. What about others, @mythi @zvonkok ?

marquiz:
Thanks @ArangoGutierrez ! I think this greatly increases the usability.

I'd be ready to merge. Has anybody really verified this change? 😉  I haven't...

marquiz:
One more comment/wish I actually have is that I think it would be good to reflect this change in the documentation (README) somehow. Even a short section about the operand configuration or smth. WDYT?

ArangoGutierrez:
> One more comment/wish I actually have is that I think it would be good to reflect this change in the documentation (README) somehow. Even a short section about the operand configuration or smth. WDYT?

More documentation never hurt anyone! sure, good idea!

manifests/0700_cr.yaml
mythi: As I wrote, my preference is just `image`. This would be consistent with what pod spec has so less confusion for the user too.

ArangoGutierrez: c9260dc

ArangoGutierrez: does c9260dc address this requested change?  @mythi

ArangoGutierrez:
All requested changes have been addressed , commits are now squashed
- @marquiz 
- @mythi

mythi (commented):
LGTM, thanks!

marquiz (approved):
Good work 👍 
/lgtm

PR #53, opened 2021-04-08
Add support for Core-Config

ArangoGutierrez:
Operand NFD Patch https://github.com/kubernetes-sigs/node-feature-discovery/pull/433  changed the way nfd-worker get's arguments, this patch looks to update the operator to the new way of running the nfd-worker 


config/manager/manager.yaml
```diff
@@ -30,7 +30,7 @@ spec:
             capabilities:
               drop: ["ALL"]
           command:
-            - /manager
+            - /node-feature-discovery-operator
```
marquiz, line 33: nit: this is another unrelated change

config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
```diff
@@ -6,10 +6,30 @@ metadata:
 spec:
   operand:
     namespace: node-feature-discovery-operator
-    image: quay.io/openshift/origin-node-feature-discovery:4.7
+    image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
```
marquiz, line 9: What image?

controllers/nodefeaturediscovery_state.go
```diff
@@ -44,8 +44,10 @@ func (n *NFD) init(
 	n.rec = r
 	n.ins = i
 	n.idx = 0
-	n.addState("/opt/nfd/master")
-	n.addState("/opt/nfd/worker")
+	if len(n.controls) == 0 {
+		n.addState("/opt/nfd/master")
+		n.addState("/opt/nfd/worker")
+	}
```
marquiz, line 50: unrelated?

config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
```diff
@@ -38,6 +38,16 @@ spec:
           spec:
             description: NodeFeatureDiscoverySpec defines the desired state of NodeFeatureDiscovery
             properties:
+              customConfig:
+                description: ConfigMap describes configuration options for the NFD
```
marquiz, line 42: Not true, this will only contain rules for the custom source

marquiz (changes requested):
@ArangoGutierrez thanks for the PR!

Generally, looks good - could you just drop the unrelated pieces and address those in separate PRs? 🥺 I think that in general one PR (or at least patch) should "do one thing" - this will make the reviewer's task a lot easier and also contribute to a more comprehensible commit history.

I also started to ponder the custom config. Those are designed for extra custom config fragments (e.g. from 3rd party) and the idea is that there can be multiple of those. I think we should either drop those at this point (custom rules can be injected directly in the main config) or make the custom configs an array (i.e. support multiple ones). WDYT?

config/manager/manager.yaml
ArangoGutierrez: https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/57

controllers/nodefeaturediscovery_state.go
ArangoGutierrez: https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/57

Makefile
```diff
@@ -1,10 +1,4 @@
-IMAGE_BUILD_CMD ?= docker build
-IMAGE_BUILD_EXTRA_OPTS ?=
-IMAGE_PUSH_CMD ?= docker push
-CONTAINER_RUN_CMD ?= docker run -u "`id -u`:`id -g`"
```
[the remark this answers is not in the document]
ArangoGutierrez: sure, going to leave `CONTAINER_RUN_CMD ?= docker run -u "`id -u`:`id -g`"` the others are repeated on line 48, that is only why I am cleaning the makefile. 
If you prefer I can add this Makefile cleanup to the #53 unrelated minor fixes PR ?

config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
ArangoGutierrez: fixed

config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
ArangoGutierrez: fixed

Makefile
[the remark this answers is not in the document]
ArangoGutierrez: fixed

ArangoGutierrez:
> an be multiple of those. I think we should either drop those at this point (custom rules can be injected directly in the main config) or make the custom configs an array (i.e. support multiple ones). WDYT?

I think that will require discussion and an extra (following) PR, the idea with operators is to always handle things via CRD's, prevent users going and creating their own config maps and injecting them into pods.  I can see how we could create more mini config maps and hold an array. but I would like to see this PR through and address the multiple custom rules in a separate PR
WDYT?

Makefile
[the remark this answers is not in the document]
marquiz: Could you submit the changes to the Makefile as a separate PR? We could/should review/merge it separately

marquiz:
> I think that will require discussion and an extra (following) PR, the idea with operators is to always handle things via CRD's, prevent users going and creating their own config maps and injecting them into pods. I can see how we could create more mini config maps and hold an array. but I would like to see this PR through and address the multiple custom rules in a separate PR
> WDYT?

I don't necessarily see the usefulness of of a single `custom-config` in the operator. Might be just a source of unnecessary confusion as the user can equally well inject the custom rules (that would go into custom-config) into the main worker config. With an array (multiple custom configs) I can see a use case of dropping custom config fragments from various sources independently, without altering any other configs. Thus, I'd suggest to drop it this for now and address it in a separate PR. How does that sound? Any thoughts from @zvonkok ?

config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 spec:
   operand:
     namespace: node-feature-discovery-operator
-    image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
+    image: gcr.io/k8s-staging-nfd/node-feature-discovery:v0.8.1
```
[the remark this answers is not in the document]
marquiz: Actually, I'd submit this change in a separate PR, too.

...and use
```
```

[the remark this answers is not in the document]
ArangoGutierrez: https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/58

Makefile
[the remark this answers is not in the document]
ArangoGutierrez: https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/59

ArangoGutierrez:
> > I think that will require discussion and an extra (following) PR, the idea with operators is to always handle things via CRD's, prevent users going and creating their own config maps and injecting them into pods. I can see how we could create more mini config maps and hold an array. but I would like to see this PR through and address the multiple custom rules in a separate PR
> > WDYT?
> 
> I don't necessarily see the usefulness of of a single `custom-config` in the operator. Might be just a source of unnecessary confusion as the user can equally well inject the custom rules (that would go into custom-config) into the main worker config. With an array (multiple custom configs) I can see a use case of dropping custom config fragments from various sources independently, without altering any other configs. Thus, I'd suggest to drop it this for now and address it in a separate PR. How does that sound? Any thoughts from @zvonkok ?

Drop `custom-config` from this PR, leave this PR to be just for `core-config` and address/discuss `custom-config` on it's won PR? makes sense to me

ArangoGutierrez:
needs rebase

Makefile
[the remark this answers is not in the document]
ArangoGutierrez: rebased

PR #55, opened 2021-04-08
Bump go version to 1.16

ArangoGutierrez:
This patch updates:
- Go to 1.16.3
- Run `go mod tidy` 


config/manager/kustomization.yaml
```diff
@@ -12,6 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: k8s.gcr.io/nfd/node-feature-discovery-operator
-  newTag: v0.2.0
-
+  newName: quay.io/eduardoarango/node-feature-discovery-operator
+  newTag: v0.2.0-22-ge57fe01f-dirty
```
marquiz, line 16: Don't change these, please

config/manager/manager.yaml
```diff
@@ -30,7 +30,7 @@ spec:
             capabilities:
               drop: ["ALL"]
           command:
-            - /manager
+            - /node-feature-discovery-operator
```
marquiz, line 33: Unrelated

marquiz (changes requested):
Remember to also submit a accompanying PR against https://github.com/kubernetes/test-infra

config/manager/kustomization.yaml
ArangoGutierrez: damm kustomize... sorry

ArangoGutierrez:
> Remember to also submit a accompanying PR against https://github.com/kubernetes/test-infra

working on that

ArangoGutierrez:
https://github.com/kubernetes/test-infra/pull/21726

config/manager/manager.yaml
ArangoGutierrez: this is fixed on #53 , had to use it here to.
after merging #53 this line won't be needed

marquiz (commented):
Sorry to nag about the commit message as it's not aligned with the content. Could you update it, e.g. to something like the PR title `Bump go version and tidy up go.sum`

ArangoGutierrez:
> Bump go version and tidy up go.sum

done, now commit has a good and descriptive message

marquiz:
> commit has a good and descriptive message

Don't see any change there 🤔

marquiz:
> Don't see any change there 🤔

Now I see it. I just think that subject of the PR ("Bump go version") would be far better subject to the commit message than the current "update dependencies". Which is quite misleading as this patch does not really udpate any dependencies

ArangoGutierrez:
> > Don't see any change there thinking
> 
> Now I see it. I just think that subject of the PR ("Bump go version") would be far better subject to the commit message than the current "update dependencies". Which is quite misleading as this patch does not really udpate any dependencies

got it

marquiz (approved):
Thanks. Now we're golang 1.16 everywhere
/lgtm

PR #56, opened 2021-04-08
add support for NFD-Master instance flag

ArangoGutierrez:
Issue https://github.com/kubernetes-sigs/node-feature-discovery/issues/427 proposed a way to handle multiple NFD instances on a same cluster , this was addressed with https://github.com/kubernetes-sigs/node-feature-discovery/pull/431 patch 

this patch updated the operator to support multiple instances of NFD on the same cluster

THIS PATCH MUST BE REVIEWED AND MERGED AFTER #53

### DIFF
--- a/build/assets/master/0400_master_daemonset.yaml
+++ b/build/assets/master/0400_master_daemonset.yaml
@@ -35,4 +35,3 @@ spec:
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
-
--- a/config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
+++ b/config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
@@ -37,6 +37,9 @@ spec:
           spec:
             description: NodeFeatureDiscoverySpec defines the desired state of NodeFeatureDiscovery
             properties:
+              instance:
+                description: Instance name. Used to separate annotation namespaces for multiple parallel deployments.
+                type: string
               operand:
                 description: OperandSpec describes configuration options for the operand
                 properties:
@@ -60,8 +63,7 @@ spec:
                     type: integer
                 type: object
               workerConfig:
-                description: ConfigMap describes configuration options for the NFD
-                  worker
+                description: WorkerConfig describes configuration options for the NFD worker
                 properties:
                   configData:
                     description: BinaryData holds the NFD configuration file
--- a/controllers/nodefeaturediscovery_controls.go
+++ b/controllers/nodefeaturediscovery_controls.go
@@ -311,12 +311,20 @@ func DaemonSet(n NFD) (ResourceStatus, error) {
 
 	// update nfd-master service port
 	if obj.ObjectMeta.Name == "nfd-master" {
+		var args []string
 		port := defaultServicePort
 		if n.ins.Spec.Operand.ServicePort != 0 {
 			port = n.ins.Spec.Operand.ServicePort
 		}
-		portFlag := fmt.Sprintf("--port=%d", port)
-		obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag}
+		args = append(args, fmt.Sprintf("--port=%d", port))
+
+		// check if running as instance
+		// https://kubernetes-sigs.github.io/node-feature-discovery/v0.8/advanced/master-commandline-reference.html#-instance
+		if n.ins.Spec.Instance != "" {
+			args = append(args, fmt.Sprintf("--instance=%s", n.ins.Spec.Instance))
+		}
+
+		obj.Spec.Template.Spec.Containers[0].Args = args
 	}
 
 	obj.SetNamespace(n.ins.GetNamespace())
--- a/api/v1/nodefeaturediscovery_types.go
+++ b/api/v1/nodefeaturediscovery_types.go
@@ -26,6 +26,7 @@ import (
 // +k8s:openapi-gen=true
 type NodeFeatureDiscoverySpec struct {
 	Operand      OperandSpec `json:"operand"`
+	Instance     string      `json:"instance"`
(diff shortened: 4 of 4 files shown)


ArangoGutierrez:
this PR buils on PR #53 so leaving it on 
/hold

config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
```diff
@@ -5,7 +5,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
-    api-approved.kubernetes.io: "unapproved, experimental-only"
+    "api-approved.kubernetes.io": "unapproved, experimental-only"
```
marquiz, line 8: Unrelated, unnecessary?

config/rbac/role.yaml
```diff
@@ -302,18 +302,3 @@ rules:
...
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - use
-  - watch
-
```
marquiz, line 319: Makes sense but could this be addressed in a separate PR (or at least patch)?

controllers/nodefeaturediscovery_controls.go
```diff
@@ -306,6 +307,22 @@ func DaemonSet(n NFD) (ResourceStatus, error) {
...
+	// https://kubernetes-sigs.github.io/node-feature-discovery/v0.8/advanced/master-commandline-reference.html#-instance
+	if n.ins.Spec.Instance != "" && obj.ObjectMeta.Name == "nfd-master" {
+		instanceFlag := fmt.Sprintf("--instance=%s", n.ins.Spec.Instance)
+		var found bool
+		for _, c := range obj.Spec.Template.Spec.Containers[0].Command {
+			if c == instanceFlag {
+				// all set!
+				found = true
+			}
+		}
```
marquiz, line 320: stupid question: can the instance flag change so that this code is unable to find it?

marquiz (changes requested):

controllers/nodefeaturediscovery_controls.go
ArangoGutierrez: Good catch!

marquiz: Hmm, could we drop this fragile-looking logic altogether and just forcefully (re-)set the container command line?

E.g. so that `build/assets/master/0400_master_daemonset.yaml` would have simply `command: ["nfd-master"]` and we would set `obj.Spec.Template.Spec.Containers[0].Args` to `{"--port=12000"}` here and append `--instance` if necessary. Would make it a lot cleaner and more robust IMO. WDYT?

marquiz (changes requested):

config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
```diff
@@ -115,4 +116,12 @@ spec:
...
-
+  customConfig:
+    configData: |
+      #    - name: "more.kernel.features"
+      #      matchOn:
+      #      - loadedKMod: ["example_kmod3"]
+      #    - name: "more.features.by.nodename"
+      #      value: customValue
+      #      matchOn:
+      #      - nodename: ["special-.*-node-.*"]
```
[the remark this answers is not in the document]
ArangoGutierrez: this patch was created/checkout on #53 git story , that's why I have it on hold. I will update once we merge the core-config patch

ArangoGutierrez:
Now this baby needs rebase

controllers/nodefeaturediscovery_controls.go
ArangoGutierrez: agree, also TODO allow users to define the master port via CRD, WDYT? 
cc @zvonkok

ArangoGutierrez:
This Patch now looks a lot better and more specific to it's intention 
ready for review @marquiz

build/assets/master/0400_master_daemonset.yaml
```diff
@@ -30,6 +30,7 @@ spec:
           image: $(NODE_FEATURE_DISCOVERY_IMAGE)
           name: nfd-master
           command: ["nfd-master", "--port=12000"]
+          args: ["--port=12000"]
```
marquiz, line 33: I think you could/should drop this from here and add it unconditionally in the controller.

controllers/nodefeaturediscovery_controls.go
marquiz: Yes of course. I'm not sure how useful that would be 🤔 Maybe it is

ArangoGutierrez:
Will rebase on #63 to unhold

ArangoGutierrez:
@marquiz this one is now ready to smash

controllers/nodefeaturediscovery_controls.go
```diff
@@ -317,6 +317,13 @@ func DaemonSet(n NFD) (ResourceStatus, error) {
 		}
 		portFlag := fmt.Sprintf("--port=%d", port)
 		obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag}
+
+		// check if running as instance
+		// https://kubernetes-sigs.github.io/node-feature-discovery/v0.8/advanced/master-commandline-reference.html#-instance
+		if n.ins.Spec.Instance != "" {
+			instanceFlag := fmt.Sprintf("--instance=%s", n.ins.Spec.Instance)
+			obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag, instanceFlag}
```
marquiz, line 325: nit: I think it would be neater and more future-proof (if we add more flags) to use `append()` here

marquiz (approved):
I like this @ArangoGutierrez! Good work 😎 

/lgtm

PR #61, opened 2021-04-16
Add kustomize patch for ClusterRole

ArangoGutierrez:
For the proper deployment of nfd-worker we need it to be provided with
specific permisions that can not be set via +kubebuilder

this patch adds a kustomize patch so we can hard code the needed rbac
rules for the porper deployment of the nfd-worker

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -1,3 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
 resources:
 - role.yaml
 - role_binding.yaml
@@ -11,3 +14,22 @@ resources:
 - auth_proxy_role_binding.yaml
 - auth_proxy_client_clusterrole.yaml
 
+# needed for nfd-worker
+# this patch is needed given that
+# +kubebuilder does not allow resourceNames
+patchesJSON6902:
+- target:
+    kind: ClusterRole
+    name: manager-role
+  patch: |-
+    - op: add
+      path: /rules/0
+      value:
+        apiGroups:
+        - policy
+        resources:
+        - podsecuritypolicies
+        verbs:
+        - use
+        resourceNames:
+        - nfd-worker


config/rbac/clusterrole_patch.yaml
```diff
@@ -0,0 +1,313 @@
...
+  verbs:
+  - watch
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
+  resourceNames:
+  - nfd-worker
```
marquiz, line 312: Could you only have this added "diff" in the patch. Not the whole content of `role.yaml`?

ArangoGutierrez: I tried to use kustomize merge strategy and it seems not to work as expected and breaks the entire yaml.
this is the only way I found via kustomize

marquiz: It can be made to work, did you try something like this:

```
rules:
- apiGroups:
  - policy
  resources:
  - podsecuritypolicies
  verbs:
  - use
  resourceNames:
  - nfd-worker
- $patch: merge
```

Json patch works for sure

ArangoGutierrez: will try and update the PR

ArangoGutierrez: Thanks for the tip, it now works, and is smaller :) and clear

config/rbac/kustomization.yaml
```diff
@@ -11,3 +14,22 @@ resources:
...
+      path: /rules/1
+      value: 
+        apiGroups:
+        - policy
+        resources:
+        - podsecuritypolicies
+        verbs:
+        - use
+        resourceNames:
+        - nfd-worker
```
marquiz, line 35: Does this work as intended? Shouldn't you add new item to rules list (`path: /rules`) instead of modifying item 1 in the list?

Also, there's trailing whitespace on some lines.

ArangoGutierrez: yup it works as expected
run `kustomize build config/default` and check the output 

trimmed the white spaces

marquiz (approved):
Thanks @ArangoGutierrez! Nice and clean 😄 

/lgtm

PR #62, opened 2021-04-16
Add kustomize rule for commonAnnotations on CRD

ArangoGutierrez:
Currently `make manifests` resets the CRD file removing a needed
annotation
- api-approved.kubernetes.io: "unapproved, experimental-only"

this patch overrides this autogeneration by adding a kustomize rule commonAnnotations into
config/crd/kustomization.yaml file, allowing kustomize to patch the crd
when `kustomize build config/default` is called, injecting the
annotation.

### DIFF
--- a/config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
+++ b/config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
@@ -5,7 +5,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
-    api-approved.kubernetes.io: "unapproved, experimental-only"
   creationTimestamp: null
   name: nodefeaturediscoveries.nfd.kubernetes.io
 spec:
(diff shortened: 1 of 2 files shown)


marquiz (commented):
Didn't work as expected. Annotation was still missing after I ran `make manifests`

ArangoGutierrez:
> Didn't work as expected. Annotation was still missing after I ran `make manifests`

I think I was short on explanation on this PR , going to edit the description

config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
```diff
@@ -5,7 +5,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
-    api-approved.kubernetes.io: "unapproved, experimental-only"
```
marquiz, line 8: Is this desired?

ArangoGutierrez: yes, is not needed any more, we can now host the default yaml we get from `make manifest` and count on `kustomize` to apply annotation when `make deploy`

marquiz (approved):
👍 

/lgtm

PR #63, opened 2021-04-16
Add serviceport var to CRD

ArangoGutierrez:
This patch allows users to modify the nfd-master Service port via CRD
with the variable Operand.ServicePort , will deafult to 12000 when not
defined

### DIFF
--- a/config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
+++ b/config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
@@ -8,6 +8,7 @@ spec:
     namespace: node-feature-discovery-operator
     image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
     imagePullPolicy: Always
+    servicePort: 12000
   workerConfig:
     configData: |
       #core:
--- a/controllers/nodefeaturediscovery_controls.go
+++ b/controllers/nodefeaturediscovery_controls.go
@@ -310,13 +310,15 @@ func DaemonSet(n NFD) (ResourceStatus, error) {
 	}
 
 	// update nfd-master service port
-	if n.ins.Spec.Operand.ServicePort != 0 && obj.ObjectMeta.Name == "nfd-master" {
-		portFlag := fmt.Sprintf("--port=%d", n.ins.Spec.Operand.ServicePort)
-		obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag}
-	} else if obj.ObjectMeta.Name == "nfd-master" {
-		portFlag := fmt.Sprintf("--port=%d", defaultServicePort)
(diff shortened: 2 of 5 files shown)


controllers/nodefeaturediscovery_controls.go
```diff
@@ -35,6 +37,8 @@ type ResourceStatus int
 const (
 	Ready ResourceStatus = iota
 	NotReady
+
+	servicePort int = 12000
```
marquiz, line 41: Suggest slight renaming:

```suggestion
	defaultServicePort int = 12000
```

config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
```diff
@@ -8,6 +8,7 @@ spec:
     namespace: node-feature-discovery-operator
     image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
     imagePullPolicy: Always
+    servicePort: 24000
```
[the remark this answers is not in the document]
ArangoGutierrez: Yup I was going for a different value to show how it works, doing something diff from default, but same as default works for me

controllers/nodefeaturediscovery_controls.go
```diff
@@ -305,6 +309,15 @@ func DaemonSet(n NFD) (ResourceStatus, error) {
...
 	}
 
+	// update nfd-master service port
+	if n.ins.Spec.Operand.ServicePort != 0 && obj.ObjectMeta.Name == "nfd-master" {
+		portFlag := fmt.Sprintf("--port=%d", n.ins.Spec.Operand.ServicePort)
+		obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag}
+	} else if obj.ObjectMeta.Name == "nfd-master" {
+		portFlag := fmt.Sprintf("--port=%d", defaultServicePort)
+		obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag}
+	}
```
marquiz, line 319: Hmm, my comment regarding this was lost somewhere yesterday 🤔  Anyway, I somehow find this more difficult to read than it should. What do you think about something like this (could probably be made polished even more but the main point is checking for "nfd-master" only once)

```suggestion
        if obj.ObjectMeta.Name == "nfd-master" {
                port := defaultServicePort
                if n.ins.Spec.Operand.ServicePort != 0 {
                        port = n.ins.Spec.Operand.ServicePort
                }
                portFlag := fmt.Sprintf("--port=%d", port)
                obj.Spec.Template.Spec.Containers[0].Args = []string{portFlag}
        }

```

marquiz (approved):
Nice improvement @ArangoGutierrez!

/label tide/merge-method-squash
/lgtm

PR #73, opened 2021-06-02
scripts/github/update-gh-pages: tolerate non-annotated tags

marquiz:
Our v0.2.0 release tag is unannotated so we need to live with that.


ArangoGutierrez (approved):
/lgtm

PR #76, opened 2021-07-13
add support for config IMAGE_TAG and rbac proxy image

zwpaper:
According to https://github.com/kubernetes-sigs/node-feature-discovery-operator/blob/b5efc6fb4b0bec190232ab47ac6fd687934d730d/docs/get-started/quick-start.md#L18

IMAGE_TAG could be configured for changing deploy image and tag, but not in the makefile, so I updated it.

Also, I add an extra variable `IMAGE_TAG_RBAC_PROXY`, ` gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0` by default, for the rbac proxy image and tag, so that we could change it when deploying.

I noticed there is both `$(shell pwd)` and `$(PROJECT_DIR)` in the Makefile, it may cause confusion, so that I changed them to a uniq `$(PROJECT_DIR)`

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
+        image: kube-rbac-proxy:latest
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
@@ -24,4 +24,3 @@ spec:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
-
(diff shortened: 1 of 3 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
+        image: kube-rbac-proxy:latest
```
ArangoGutierrez, line 13: Let's keep this one as `v0.5.0` avoid using `latest` tag at all cost

ArangoGutierrez (changes requested):
Thanks for this PR @zwpaper 
just a comment before moving forward

config/default/manager_auth_proxy_patch.yaml
zwpaper: updated to `v0.5.0`, although we will not use this image and tag here, it should always be updated to the one in Makefile by kustomization.

I noticed one of my mistake, kustomization should change the RBAC image in `config/default` instead of `config/manager`, it is now tested and works as expected.

btw, I also noticed there is both `$(shell pwd)` and `$(PROJECT_DIR)` in the Makefile, it may cause confusion, so that I changed them to a uniq `$(PROJECT_DIR)`

ArangoGutierrez (approved):
Thanks for the patch! 
/lgtm

PR #78, opened 2021-08-04
Add BASE_IMAGE_FULL build arg

ArangoGutierrez:
This patch :
- Bumps base go image to 1.16.6
- updates second stage image to be consistent with NFD via build args

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-# Build the manager binary
+ARG BASE_IMAGE_FULL
+# Build the manager biinary
 FROM golang:1.16.6-buster as builder
 
 WORKDIR /workspace
@@ -14,7 +15,7 @@ COPY . .
 RUN make build
 
 # Create production image for running the operator
-FROM registry.access.redhat.com/ubi8/ubi
+FROM ${BASE_IMAGE_FULL}
 COPY --from=builder /workspace/node-feature-discovery-operator /
 
 RUN mkdir -p /opt/nfd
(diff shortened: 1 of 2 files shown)


ArangoGutierrez:
> Looks basically good. Just pondering the naming of the variable, are you planning to have a `minimal` image similar to NFD (operand)? If not, we could drop the `_FULL` suffix from it.

https://github.com/ArangoGutierrez/node-feature-discovery-operator/commit/642bf6a351456622cda9e69fbbac3736640411f2

marquiz:
nit: the title of this PR is misleading. I'd change it to something like **Add BASE_IMAGE_FULL build arg**

marquiz (commented):
I think we'd better add the minimal image in a separate PR.

Need to add CI support (build-image and push-image) plus preferably mention that in the documentation, too.

Talking about CI, should we have an `images` Makefile target building both images and use that in `scripts/test-infra/build-image.sh` 🤔

ArangoGutierrez:
> I think we'd better add the minimal image in a separate PR.
> 
> Need to add CI support (build-image and push-image) plus preferably mention that in the documentation, too.
> 
> Talking about CI, should we have an `images` Makefile target building both images and use that in `scripts/test-infra/build-image.sh` thinking

OK, removed the commit, will turn that one into a PR, + documentation + CI (in a single PR)

marquiz (approved):
Looks good, now

/lgtm

ArangoGutierrez:
Follow up PR https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/82

PR #77, opened 2021-07-20
add support for extra-label-ns

zwpaper:
NFD validates label ns by `--extra-label-ns` args, but the operator has not yet supported it.

this PR added an `extraLabelNS` as a list of strings under `Operand` for the support

PR #79, opened 2021-08-05
rbac: grant node/status update permission to manager and master

zwpaper:
after https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/77 merged, NFD operator can create resources labels, it will need this RBAC permission to update the node resources.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -105,6 +104,14 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - nodes/status
+  verbs:
+  - get
+  - patch
+  - update
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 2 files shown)


config/rbac/role.yaml
```diff
@@ -105,6 +104,14 @@ rules:
...
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - nodes/status
+  verbs:
+  - get
+  - patch
+  - update
```
marquiz, line 114: Hmm, why would the manager need the permission to update node status?

zwpaper: hi @marquiz , there is a feature in NFD adding some labels as `resource-label`, NFD operator will support it after https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/77 merged, the `resource-label` means some 3rd party resource for the node, which is exposed in `status.allocatable` and `status.capacity` so that this permission would be needed.

zwpaper: Oops, sorry for not realized you were talking about the manager. 
because the role that the operand used was created by the manager, and the manger can not grant the node/status if it did not have it.

marquiz: OK, thx for the explanation. I wasn't aware of that restriction

PR #80, opened 2021-08-09
Use apimachinery error library to check for missing nfd instance

courtneypacheco:
The existing error checking library in the cri-api package was used to check for the case where the NFD instance is deleted. However, this error checking library is used for GRPC specific errors, not for errors related to the `metav1` package. The `metav1` package is used by various k8s resources (e.g., DaemonSet, Service, etc.), and thus, we should be checking for `metav1` problems, not GRPC problems.

### DIFF
--- a/controllers/nodefeaturediscovery_controller.go
+++ b/controllers/nodefeaturediscovery_controller.go
@@ -22,9 +22,9 @@ import (
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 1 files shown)


marquiz:
Thanks @courtneypacheco for the patch!

Need to gofmt:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_node-feature-discovery-operator/80/pull-node-feature-discovery-operator-verify/1424869545398505472

Otherwise lgtm

ArangoGutierrez:
Thanks for the review @marquiz 
/approve

marquiz:
nit: could you squash the commits into one? gofmt should be part of the first (single) patch.

courtneypacheco:
Thanks, @marquiz! I'm ready for review again whenever you have the time!

marquiz (approved):
Thanks @courtneypacheco! Looks good. Plus I really like your commit messages 😊 

/lgtm

PR #84, opened 2021-08-26
Add logic to set the operator's status in the NFD CR

courtneypacheco:
Add logic that sets the operator's status in the NFD CR based on whether one or more of NFD's resources is: degraded, progressing, upgradeable, or available. Also add a "applyComponents" function to simplify readability.

### DIFF
--- a/controllers/nodefeaturediscovery_status.go
+++ b/controllers/nodefeaturediscovery_status.go
@@ -1,0 +1,567 @@
+package controllers
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	nfdv1 "github.com/kubernetes-sigs/node-feature-discovery-operator/api/v1"
+)
+
+// nodeType is either 'worker' or 'master'
+type nodeType int
+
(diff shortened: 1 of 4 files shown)


courtneypacheco:
@marquiz Thanks! I noticed some of my tests failed, so I'm adding fixes. I will certainly rebase. :)

courtneypacheco:
@marquiz Just squashed commits and rebased. PTAL.

ArangoGutierrez:
> @marquiz Just squashed commits and rebased. PTAL.

@marquiz lgtm?

ArangoGutierrez (approved):
This works downstream
waiting on @marquiz for lgtm

controllers/nodefeaturediscovery_status.go
```diff
@@ -0,0 +1,567 @@
+package controllers
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
```
marquiz, line 8: Hmm, I'm not sure we want to depend on downstream (openshift) at least this heavily 🧐   I think we should use k8s metav1 (`k8s.io/apimachinery/pkg/apis/meta/v1`).

Maybe `ConditionType`s is something that we could use from openshift 🤔  Or then just copy paste those here in order to break the dependency

marquiz (changes requested):
Thanks for the PR!

I gave it a quick look, not ready for merging, yet 😉  I'm most concerned about the dependency to openshift. I know we already have some but we should keep it at a very minimum (and possibly eliminate some of the existing, too)

ArangoGutierrez:
Superseed by https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/88
/close

PR #67, opened 2021-05-14
Add documentation to the controller code and main.go

courtneypacheco:
Adding documentation to the controller code and main.go so that contributors and users have an easy time understanding the purpose of the functions, variables, structs, etc. in the codebase.

PR #87, opened 2021-09-28
Enhance in source documentation

ArangoGutierrez:
This is a re-write of #67 

Adding documentation to the code in 'main.go' so that contributors
and developers can understand the purpose of each function, variable,
etc..

Adding documentation to funcs and vars in nodefeaturediscovery_controller.go

Adding documentation to various functions and variables within the
nodefeaturediscovery_controller.go file so that users and contributors
can have a deeper understanding of how the reconciliation process
works with the NFD Operator.

Adding documentation for NFD controls functions

Adding documentation to describe what each of the vars and funcs in
the NFD controls file does, when to use them, and why they're
important.

Adding documentation to the NFD controller resources file

Adding documentation describing the funcs, vars, etc. in the NFD
controller resources file so that users and contributors can
understand how they all work.

Adding docs to NFD state funcs related to NFD itself (not just the operator)

Adding documentation to the NFD state functions related to NFD itself
so that users and contributors can understand how NFD works with the NFD
operator, especially if they are looking at another file that references
these functions and the NFD struct.

fix gofmt formatting

Signed-off-by: Courtney Pacheco <cpacheco@redhat.com>

Shorten controller and main.go documentation

The documentation in the controller and main.go files was
unnecessarily long and could pose problems in the future if a
maintainer or contributor wants to modify a feature that has a lot
of existing documentation. Ideally, documentation should be concise.

Update NFD documentation based on feedback

Update the NFD documentation based on feedback in the PR. Will
squash commits before merging.

### DIFF
--- a/controllers/nodefeaturediscovery_controller.go
+++ b/controllers/nodefeaturediscovery_controller.go
@@ -37,29 +37,51 @@ import (
 	nfdv1 "github.com/kubernetes-sigs/node-feature-discovery-operator/api/v1"
 )
 
+// log is used to set the logger with a name that describes the actions of
+// functions and arguments in this file
 var log = logf.Log.WithName("controller_nodefeaturediscovery")
 
+// nfd is an NFD object that will be used to initialize the NFD operator
 var nfd NFD
 
 // NodeFeatureDiscoveryReconciler reconciles a NodeFeatureDiscovery object
 type NodeFeatureDiscoveryReconciler struct {
+
+	// Client interface to communicate with the API server. Reconciler needs this for
+	// fetching objects.
 	client.Client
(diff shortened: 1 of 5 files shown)


marquiz (changes requested):
Change looks good 👍  The commit message looks horrible 🙀  😄 

Could you tidy up the commit message. It think in this case even the subject alone (probably `Enhance in-source documentation`?), without any body would be enough.

ArangoGutierrez:
@marquiz commit fixed

marquiz (approved):
Thanks @ArangoGutierrez, good to have this finally merged 👍 
/lgtm

PR #89, opened 2021-09-30
Clean up SCC OpenShift deps

ArangoGutierrez:


marquiz (commented):
Cool @ArangoGutierrez, good cleanup.

Should we remove `security.openshift.io` from `config/rbac/` roles, too?

marquiz:
Oh, and should the commit message subject be *deps* instead of *depts*?

marquiz (approved):
Nice, I like reducing LOC 😄 
/lgtm

PR #90, opened 2021-09-30
Drop pkq folder

ArangoGutierrez:
Folder pkg is named pkq with Q, this patch addresses that.
Also clean non needed lines of code 


marquiz (approved):
Nice catch of dead code
/lgtm

PR #108, opened 2022-01-19
release-0.2: docs: drop invalid step from deployment instructions

marquiz:


ArangoGutierrez (approved):
This is correct, the Makefile rules go all the way to deploying the sample CR, my bad, on downstream I am not, this is why I got it wrong on the documentation 
/lgtm

PR #109, opened 2022-01-20
docs: link to correct version of operand docs

marquiz:
Change the sidebar link to NFD (operand) documentation to point to the
version fixed in the docs config (and not automatically to the latest
release). This makes sure that docs of older releases point to an
operand version that they were tested with.


ArangoGutierrez (approved):
/lgtm

PR #138, opened 2022-04-11
Use runtime/debug to extract VCS information

qbarrand:
This PR leverages the `runtime/debug` package to extract VCS build settings from the binary.
It mimicks the old behaviour - only the number of commits since the last tag is not included anymore, as that information is not available as a build setting.


marquiz:
Hi @qbarrand. What is the benefit/use case pursued with this PR?

It doesn't work in backwards-compatible manner:

```
(pr-138)$ ./node-feature-discovery-operator --version
(pr-138)$ make build VERSION=v1.2.3
nfd-operator v1.2.3-2046e655e3059b8a08dba67120cfda69bd2ba99e-dirty
```

With `master` we get:

```
(master)$ make build VERSION=v1.2.3
(master)$ ./node-feature-discovery-operator --version
nfd-operator v1.2.3
```

We rely on this behavior (`VERSION` setting the exact version reported by the binary) in CI for example

Also, let's bump to go v1.18 in a separate PR.

qbarrand:
Hi @marquiz

> What is the benefit/use case pursued with this PR?

@ArangoGutierrez asked me to implement this in NFD as we are moving to this Go 1.18 feature instead of ldflags in other operators.

> It doesn't work in backwards-compatible manner:

Indeed. I was assuming that `VERSION` in `Makefile` was always used - @ArangoGutierrez can you clarify what behaviour is desirable here?
I tried to replicate what `VERSION` was doing, but I would rather always write the VCS build settings to the logs at startup regardless of whether `-version` was provided or not. IMHO that can be valuable information when troubleshooting, and we could leave the `VERSION` mechanism unchanged.

PR #151, opened 2022-04-28
controller: fix nfd-master cmdline args

marquiz:
Remove quoting of the arg values which effectively make all of them
broken.


ArangoGutierrez (approved):
/lgtm

PR #155, opened 2022-05-10
Minor updates to kustomize set image

ArangoGutierrez:


.gitignore
```diff
@@ -23,3 +23,4 @@ testbin/*
 *.swp
 *.swo
 *~
+.vscode/*
```
marquiz, line 26: nit: how this relates to kustomize?

marquiz (commented):
We do a lot of small (independent?) changes in this PR and at a quick glance it's a bit hard to follow what/why. I'd suggest splitting the independent/orthogonal changes into separate commits or at least elaborate a bit in the commit message (what was done and why)

.gitignore
ArangoGutierrez: it doesn't but my vscode decided to become verbose recently, and is outputting files and logs there. if you prefer I can drop that line from this PR

ArangoGutierrez: going to split every change of this PR into mini tiny commits 
this is a nit PR mostly, cosmetic changes here and there

ArangoGutierrez: Now every misselaneous change lives on it's own commit

config/default/kustomization.yaml
```diff
@@ -1,7 +1,7 @@
 images:
 - name: kube-rbac-proxy
   newName: gcr.io/kubebuilder/kube-rbac-proxy
-  newTag: v0.5.0
+  newTag: v0.8.0
```
marquiz, line 4: Sorry to nit: is this intended? Related to prometheus?

ArangoGutierrez: moved to a separate commit

PR #157, opened 2022-05-17
Swap default image to Minimal 

### LINKED ISSUE #146: Swap default image to Minimal
Currently the default image is the normal tag, moving forward it would be more convenient to have minimal as the default and normal as the option.

ArangoGutierrez:
Fixes #146

### DIFF
--- a/config/manifests/bases/nfd-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/nfd-operator.clusterserviceversion.yaml
@@ -13,7 +13,7 @@ metadata:
           },
           "spec": {
             "operand": {
-              "image": "gcr.io/k8s-staging-nfd/node-feature-discovery:master",
+              "image": "gcr.io/k8s-staging-nfd/node-feature-discovery:master-minimal",
               "imagePullPolicy": "Always",
               "servicePort": 12000
             },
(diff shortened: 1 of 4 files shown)


config/manifests/bases/nfd-operator.clusterserviceversion.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
           },
           "spec": {
             "operand": {
-              "image": "gcr.io/k8s-staging-nfd/node-feature-discovery:master",
+              "image": "k8s.gcr.io/nfd/node-feature-discovery:v0.11.0-minimal",
```
marquiz, line 16: Think we should use master-minimal here.

marquiz (commented):
Thanks @ArangoGutierrez. Does documentation need any updating?

config/manifests/bases/nfd-operator.clusterserviceversion.yaml
ArangoGutierrez: ok

config/manifests/bases/nfd-operator.clusterserviceversion.yaml
```diff
@@ -463,8 +463,8 @@ spec:
                 - name: OPERATOR_NAME
                   value: cluster-nfd-operator
                 - name: NODE_FEATURE_DISCOVERY_IMAGE
-                  value: k8s.gcr.io/nfd/node-feature-discovery:master
-                image: k8s.gcr.io/nfd/node-feature-discovery-operator:master
+                  value: k8s.gcr.io/nfd/node-feature-discovery:v0.11.0-minimal
```
marquiz, line 466: Needs to be master, too
```suggestion
                  value: k8s.gcr.io/nfd/node-feature-discovery:master-minimal
```

ArangoGutierrez: This registry doesn't hold a master tag. Only staging has a master tag.
Will make the change and push

PR #160, opened 2022-05-23
docs: update nfd-worker sample config

marquiz:
Sync the example from operand repository
(deployment/components/worker-config/nfd-worker.conf.example).


marquiz:
ping @ArangoGutierrez

PR #171, opened 2023-01-16
Deprecation of package ioutil

ArangoGutierrez:
"io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. (SA1019)go-staticcheck

https://github.com/go-critic/go-critic/issues/1019


marquiz (approved):
Thanks @ArangoGutierrez 
/lgtm

PR #173, opened 2023-01-17
dockerfile: builder image configurable through a Makefile variable.

ArangoGutierrez:
Switch over to the "non-point-release" version of the image. Now we always use the latest patch version of golang with latest security fixes, for example, without the need to manually bump the version after every point release.

This patch also makes the builder image configurable through a Makefile variable pointing to the go.mod go version, that way the go.mod file becomes the source of truth for the go version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 
 GO_CMD ?= go
 GO_FMT ?= gofmt
+GO_VERSION := $(shell awk '/go /{print $$2}' go.mod|grep -v v)
 CONTAINER_RUN_CMD ?= docker run -u "`id -u`:`id -g`"
 
 # Docker base command for working with html documentation.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,6 +4,7 @@
 
 GO_CMD ?= go
 GO_FMT ?= gofmt
+GO_VERSION := $(shell awk '/go /{print $$2}' go.mod|grep -v v)
```
marquiz, line 7: This is a bit hack'ish and breaks easily I think as it can return multiple lines of data. I'd suggest to harden it a bit by making sure that we only return the first matching line (we know the go version is specified in the beginning of the file) and also requiring that the string `go` is in the beginning of the line.

```suggestion
GO_VERSION := $(shell awk '/^go /{print $$2}' go.mod|head -n1)
```


 We could also get the version with 
```
go list -m -f '{{.GoVersion}}'
```

But I guess that's not so nice as it makes the build dependent on some (recent enough) golang version being installed on the host 🤔

marquiz (commented):
Thanks @ArangoGutierrez! Makes sense, only one suggestion regarding the shell script snippet for parsing the go version from go.mod.

marquiz (approved):
Thanks @ArangoGutierrez, looks good to me
/lgtm

PR #172, opened 2023-01-16
Update README to be more welcoming 

ArangoGutierrez:
This patch:
- Updates the readme making it more welcoming and providing quick start instructions

PR #178, opened 2023-01-30
README: fix documentation link

ArangoGutierrez:
FIx wrong link introduced in #172

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ kubectl apply -f config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
 
 For more detailed information on how to use the Node Feature Discovery operator,
 please check out our 
-[documentation](https://github.com/kubernetes-sigs/node-feature-discovery-operator/blob/master/docs/index.md)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ kubectl apply -f config/samples/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
 
 For more detailed information on how to use the Node Feature Discovery operator,
 please check out our 
-[documentation](https://github.com/kubernetes-sigs/node-feature-discovery-operator/blob/master/docs/index.md)
+[documentation](https://kubernetes-sigs.github.io/node-feature-discovery-operator/master/get-started/index.html)
```
marquiz, line 52: How about just referencing the docs root dir (and not specific html page) and let our redirection get the user to the correct page? That way we should avoid breakage in case docs file/directory structure changes.

```suggestion
[documentation](https://kubernetes-sigs.github.io/node-feature-discovery-operator/master)
```

marquiz (approved):
/lgtm

PR #183, opened 2023-03-10
scripts/test-infra: bump golangci-lint to v1.51.2

marquiz:

PR #184, opened 2023-03-10
Fix gofmt errors

marquiz:
Golang v1.20.2 is stricter than v1.20.


marquiz:
Needed to include this in #183 
/close

PR #185, opened 2023-03-13
HELM: update-gh-pages.sh to create chart index

ArangoGutierrez:
Add function update_helm_repo_index()

### DIFF
--- a/scripts/github/update-gh-pages.sh
+++ b/scripts/github/update-gh-pages.sh
@@ -22,9 +22,59 @@ create_versions_js() {
     # 'stable' is a symlink pointing to the latest version
     [ -f stable ] && echo "    { name: 'stable', url: '$_baseurl/stable' },"
     for f in `ls -d */  | tr -d /` ; do
-        echo "    { name: '$f', url: '$_baseurl/$f' },"
+        if [ -f "$f/index.html" ]; then
+            echo "    { name: '$f', url: '$_baseurl/$f' },"
+        fi
     done
     echo -e "  ];\n}"
+}
+
+# Helper for updating help repo index
+update_helm_repo_index() {
+    echo "Updating Helm repo index"
+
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
Tested locally: 
```bash
[eduardo@fedora node-feature-discovery-operator]$ helm show chart nfd-o/nfd-operator 
apiVersion: v2
appVersion: master
description: 'The Node Feature Discovery Operator is a Kubernetes operator that deploys
  and  manages the Node Feature Discovery (NFD) lifecycle. '
home: https://github.com/kubernetes-sigs/node-feature-discovery-operator
keywords:
- feature-discovery
- feature-detection
- node-labels
name: nfd-operator
sources:
- https://github.com/kubernetes-sigs/node-feature-discovery-operator
type: application
version: 0.1.0
```

scripts/github/update-gh-pages.sh
```diff
@@ -161,11 +214,11 @@ echo "Committing changes..."
...
 
-cd -
+popd > /dev/null
 
 echo "gh-pages branch successfully updated"
 
 if [ -n "$push_remote" ]; then
     echo "Pushing gh-pages to $push_remote"
-    git push ${amend+-f} "$push_remote" gh-pages
+    git push --force ${amend+-f} "$push_remote" gh-pages
```
marquiz, line 223: I'm not sure we want to `--force`. If the push fails (because of git history doesn't apply) we have some problem we better get notified and investigate the issue

ArangoGutierrez: removing.... it was for devel forgot to rm

marquiz (approved):
Great addition @ArangoGutierrez 🎉 Looks fine to me. Let's fix things later if we hit any problems 😀

marquiz:
Oh, and let's get this merged (as soon as prox gets back in shape...)
/lgtm

PR #191, opened 2023-03-20
Bump release to v0.6.0

ArangoGutierrez:


docs/_config.yml
```diff
@@ -42,15 +42,15 @@ remote_theme: rundocs/jekyll-rtd-theme@v2.0.10
...
 # in the sidebar (top left corner of the page)
-release: master
+release: v0.6.0
 
 # Container image which to point to in the documentation
-container_image: gcr.io/k8s-staging-nfd/node-feature-discovery-operator:master-minimal
+container_image: registry.k8s.io/nfd/node-feature-discovery-operator:v0.6.0
 # Operand Node Feature Discovery documentation version
-operand_version: master
+operand_version: v0.12.1
```
marquiz, line 56: IIRC we made the same mistake last time 😎. We should probably update the comment that this should be the "release branch version" or something, I don't know how to put it

```suggestion
operand_version: v0.12
```

config/manifests/bases/nfd-operator.clusterserviceversion.yaml
```diff
@@ -463,8 +463,8 @@ spec:
                 - name: OPERATOR_NAME
                   value: cluster-nfd-operator
                 - name: NODE_FEATURE_DISCOVERY_IMAGE
-                  value: gcr.io/k8s-staging-nfd/node-feature-discovery:master-minimal
-                image: gcr.io/k8s-staging-nfd/node-feature-discovery-operator:master-minimal
+                  value: registry.k8s.io/nfd/node-feature-discovery:v0.12.1
+                image: registry.k8s.io/nfd/node-feature-discovery-operator:v0.6.0
```
marquiz, line 467: There's another reference to `registry.k8s.io/nfd/node-feature-discovery-operator:master` above in the annotations, do we want to change that, too?

Also, I've probably asked this before (but forgot the answer), do we want to update `bundle/manifests/nfd-operator.clusterserviceversion.yaml` similarly? Also in that file (under bundle/) there are strange references e.g. to `nfd-operator.v0.5.0`, https://kubernetes-sigs.github.io/node-feature-discovery-operator/v0.4. Should/could we fix those too?

There's also a strange reference to  https://kubernetes-sigs.github.io/node-feature-discovery/v0.10/get-started/introduction.html#nfd-topology-updater above in this file. We might want to update that (in the master branch, too to point to _stable_)

marquiz (approved):
Thanks @ArangoGutierrez, looks good to me
/lgtm

Just a side note that we probably want to fix some of those outdated references in master branch as well

PR #196, opened 2023-03-31
Rename images

marquiz:
Remove the -minimal suffix from the minimal container image tag, making
it the default. Also change the naming to call it the "prod" image
internally. Rename the formerly full (suffix-less) image to "debug" to
indicate that it is really only intended for development/debugging
purposes.

This patch also stops publishing the debug images in the k8s staging
container image repository. We still test the buildability, though.


ArangoGutierrez (approved):
/lgtm

PR #202, opened 2023-10-04
Don't create service if NodeFeature API is enabled

ArangoGutierrez:


config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
```diff
@@ -45,6 +44,11 @@ spec:
                   type: string
                 nullable: true
                 type: array
+              grpcMode:
+                description: GrpcMode defines whether the NFD-master should be deployed
+                  in gRPC mode or not. If enabled, the Operator will deploy an NFD-Master
+                  in gRPC mode that will listen for incoming gRPC requests.
```
marquiz, line 50: Should we describe what happens if gRPC mode is disabled? I.e. the NodeFeature CRD API is disabled.

controllers/nodefeaturediscovery_controls.go
```diff
@@ -451,18 +451,21 @@ func Deployment(n NFD) (ResourceStatus, error) {
...
+		if n.ins.Spec.Operand.ServicePort != 0 {
+			port = n.ins.Spec.Operand.ServicePort
+		}
 
-	// Now that the port has been determined, append it to
-	// the list of args
-	args = append(args, fmt.Sprintf("--port=%d", port))
+		// Now that the port has been determined, append it to
+		// the list of args
+		args = append(args, fmt.Sprintf("--port=%d", port))
```
marquiz, line 467: We should also specify `-enable-nodefeature-api={true|false}` based on GrpcMode, right? And for both nfd-master and nfd-worker.

nit: one dash is enough 😜
```suggestion
		args = append(args, fmt.Sprintf("-port=%d", port))
```

marquiz (commented):
Thanks @ArangoGutierrez for the enhancement. Quick comments below
/cc @yevgeny-shnaidman

marquiz:
Also, nit about the commit msg title. Start with capital letter and shorten a bit, e.g.
`Don't create service if NodeFeature API is enabled`

yevgeny-shnaidman:
looks good to me

marquiz:
/retitle Don't create service if NodeFeature API is enabled

ArangoGutierrez:
/retitle Don't create service if NodeFeature API is enabled

controllers/nodefeaturediscovery_controls.go
```diff
@@ -451,18 +451,24 @@ func Deployment(n NFD) (ResourceStatus, error) {
...
 
-	// If the operand service port has already been defined,
-	// then set "port" to the defined port. Otherwise, it is
-	// ok to just use the defaultServicePort value
-	if n.ins.Spec.Operand.ServicePort != 0 {
-		port = n.ins.Spec.Operand.ServicePort
-	}
+	if n.ins.Spec.GrpcMode && obj.Name == nfdMasterApp {
+		// Disable NodeFeature API
+		args = append(args, "-enable-nodefeature-api=false")
```
marquiz, line 457: Should we correspondingly have `-enable-nodefeature-api=true` for the non-GrpcMode? So that also v0.13 of the operand could be used?

ArangoGutierrez: 🤔  next Operator release will probably be for v0.14 which comes with the API enabled by default. so... I am open to suggestions. If you think we should then I will

marquiz: I don't have a clear answer. Is the operator supposed to support some version skew? Or is it taken that the operator only works with one version of the operand 🤷

marquiz: Ah, but we have the nfd-gc as well in NFD v0.14 and possibly other breakages 🧐 Maybe the operator and operand just need to go in lock-step wrt versions

ArangoGutierrez: Yup, I see the operator locked to specific NFD versions, mostly for the hard coded (Opinionated) flags and values set by the operator

PR #203, opened 2023-10-04
Add Tolerations on NFD CR

ArangoGutierrez:


api/v1/nodefeaturediscovery_types.go
```diff
@@ -87,8 +87,14 @@ type OperandSpec struct {
 
 	// ServicePort specifies the TCP port that nfd-master
 	// listens for incoming requests.
+	// DEPRECATED: gRPC is being deprecated in favor of the NodeFeature API
+	// This will be removed in a future release
```
marquiz, line 91: nit: we might want make this deprecation in a separate PR to make it more explicit and clearer

marquiz (commented):
Thx @ArangoGutierrez. Quick comments below
/cc @yevgeny-shnaidman

api/v1/nodefeaturediscovery_types.go
```diff
@@ -87,8 +87,14 @@ type OperandSpec struct {
...
 	// +kubebuilder:validation:Optional
 	ServicePort int `json:"servicePort"`
+
+	// Toleration defines the toleration for the NFD pods
+	// +kubebuilder:validation:Optional
+	Tolerations []corev1.Toleration `json:"toleration,omitempty"`
```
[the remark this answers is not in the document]
ArangoGutierrez: I mean, for the Slack proposed use case, this will solve it. Unless users explicit want master and worker to not be on the same node bc some reason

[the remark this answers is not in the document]
yevgeny-shnaidman: Do we want to add also nodeSelector field? This will give user additional control over where worker pods should be running...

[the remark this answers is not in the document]
ArangoGutierrez: NodeSelector is better as a follow up PR to keep this one clean

config/crd/bases/nfd.kubernetes.io_nodefeaturediscoveries.yaml
```diff
@@ -4,7 +4,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.8.0
-    api-approved.kubernetes.io: unapproved, experimental-only
```
marquiz, line 7: How does this annotation get removed? Can we get it back in? 🧐

ArangoGutierrez: `make generate manifest` did this

marquiz: Hmm, ok, can we prevent that auto-generation from doing this? In a follow-up PR that is?

PR #209, opened 2023-10-11
Move to Operator-Framework Helm based

ArangoGutierrez:
Based on https://sdk.operatorframework.io/docs/building-operators/helm/


ArangoGutierrez:
> @ArangoGutierrez: The following tests **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-node-feature-discovery-operator-verify	[1bdcedc](https://github.com/kubernetes-sigs/node-feature-discovery-operator/commit/1bdcedc3e14c599f465765dfce6170235fc932ed)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_node-feature-discovery-operator/209/pull-node-feature-discovery-operator-verify/1712118233483448320)	true	`/test pull-node-feature-discovery-operator-verify`
> pull-node-feature-discovery-operator-verify-docs	[1bdcedc](https://github.com/kubernetes-sigs/node-feature-discovery-operator/commit/1bdcedc3e14c599f465765dfce6170235fc932ed)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_node-feature-discovery-operator/209/pull-node-feature-discovery-operator-verify-docs/1712118233533779968)	true	`/test pull-node-feature-discovery-operator-verify-docs`
> pull-node-feature-discovery-operator-build-image	[1bdcedc](https://github.com/kubernetes-sigs/node-feature-discovery-operator/commit/1bdcedc3e14c599f465765dfce6170235fc932ed)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_node-feature-discovery-operator/209/pull-node-feature-discovery-operator-build-image/1712118233584111616)	true	`/test pull-node-feature-discovery-operator-build-image`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=node-feature-discovery-operator&pr=209). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3AArangoGutierrez). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/node-feature-discovery-operator/issues?q=is:issue+is:open) when you hit one in your PR.

/hold
/needs-ok-to-test

ArangoGutierrez:
This PR proposes moving from a GO based operator to a HELM based operator 
NFD has become very good at being deployed by HELM, and from an operator point of view we only need a controller to keep objects healthy.
The operator has been lagging behind NFD development, with this move, NFD Helm chart becomes the CRD it self, making it easier to maintain the Operator, and also makes the operator more flexible to control NFD deployment.

yevgeny-shnaidman:
> This PR proposes moving from a GO based operator to a HELM based operator NFD has become very good at being deployed by HELM, and from an operator point of view we only need a controller to keep objects healthy. The operator has been lagging behind NFD development, with this move, NFD Helm chart becomes the CRD it self, making it easier to maintain the Operator, and also makes the operator more flexible to control NFD deployment.

I am not very familiar with help operator, but i have a couple of questions:
1) in the current operator we have a finalization logic, would we need it in helm-operator?
2) how is it going to handle prune option deployment during NFD CR deletion?
3) moving to helm operator removes a lot of code, but on the other hand it limits our flexibility, in case we want to add a new feature to NFD operator

ArangoGutierrez:
> > This PR proposes moving from a GO based operator to a HELM based operator NFD has become very good at being deployed by HELM, and from an operator point of view we only need a controller to keep objects healthy. The operator has been lagging behind NFD development, with this move, NFD Helm chart becomes the CRD it self, making it easier to maintain the Operator, and also makes the operator more flexible to control NFD deployment.
> 
> I am not very familiar with help operator, but i have a couple of questions:
> 
> 1. in the current operator we have a finalization logic, would we need it in helm-operator?
> 2. how is it going to handle prune option deployment during NFD CR deletion?
> 3. moving to helm operator removes a lot of code, but on the other hand it limits our flexibility, in case we want to add a new feature to NFD operator

1. It will have finalizers, just test it and see. 
2. Prune is a NFD option, not an Operator option. so You can pass that via Helm values on the CR.
3.  I don't see how. I think this moves the operator to a more healthy state. it has been un touched for a long time, with this, Maintaining the NFD-Operator becomes part of maintaining NFD Helm Charts.  IF we need new features they should go into NFD, the operator is just to control the lifecycle of the Operand.

yevgeny-shnaidman:
> > > This PR proposes moving from a GO based operator to a HELM based operator NFD has become very good at being deployed by HELM, and from an operator point of view we only need a controller to keep objects healthy. The operator has been lagging behind NFD development, with this move, NFD Helm chart becomes the CRD it self, making it easier to maintain the Operator, and also makes the operator more flexible to control NFD deployment.
> > 
> > 
> > I am not very familiar with help operator, but i have a couple of questions:
> > 
> > 1. in the current operator we have a finalization logic, would we need it in helm-operator?
> > 2. how is it going to handle prune option deployment during NFD CR deletion?
> > 3. moving to helm operator removes a lot of code, but on the other hand it limits our flexibility, in case we want to add a new feature to NFD operator
> 
> 1. It will have finalizers, just test it and see.
> 2. Prune is a NFD option, not an Operator option. so You can pass that via Helm values on the CR.
> 3. I don't see how. I think this moves the operator to a more healthy state. it has been un touched for a long time, with this, Maintaining the NFD-Operator becomes part of maintaining NFD Helm Charts.  IF we need new features they should go into NFD, the operator is just to control the lifecycle of the Operand.

Regarding 3:  new features like tolerations and node selectors are operator features. So they either go into the Go code, or we need to set them in the Helm charts based on NFD CR.

ArangoGutierrez:
> > > > This PR proposes moving from a GO based operator to a HELM based operator NFD has become very good at being deployed by HELM, and from an operator point of view we only need a controller to keep objects healthy. The operator has been lagging behind NFD development, with this move, NFD Helm chart becomes the CRD it self, making it easier to maintain the Operator, and also makes the operator more flexible to control NFD deployment.
> > > 
> > > 
> > > I am not very familiar with help operator, but i have a couple of questions:
> > > 
> > > 1. in the current operator we have a finalization logic, would we need it in helm-operator?
> > > 2. how is it going to handle prune option deployment during NFD CR deletion?
> > > 3. moving to helm operator removes a lot of code, but on the other hand it limits our flexibility, in case we want to add a new feature to NFD operator
> > 
> > 
> > 
> > 1. It will have finalizers, just test it and see.
> > 2. Prune is a NFD option, not an Operator option. so You can pass that via Helm values on the CR.
> > 3. I don't see how. I think this moves the operator to a more healthy state. it has been un touched for a long time, with this, Maintaining the NFD-Operator becomes part of maintaining NFD Helm Charts.  IF we need new features they should go into NFD, the operator is just to control the lifecycle of the Operand.
> 
> Regarding 3: new features like tolerations and node selectors are operator features. So they either go into the Go code, or we need to set them in the Helm charts based on NFD CR.

Agree, we are currently working to get those options into NFD deployment(Helm) 
I have an open PR with Automated Taints under review, so the idea is every functionality goes into NFD, and the operator is to manage it's life cycle, not to add new features

ArangoGutierrez:
Gently reminder PING
@marquiz @yevgeny-shnaidman

ArangoGutierrez:
@marquiz @yevgeny-shnaidman @zvonkok 
Hope we can take a look this week at this proposal

marquiz (commented):
Thanks @ArangoGutierrez. I think this would make sense.

With this – the operator basically going hand-in-hand-lockstep with the Helm chart - the immediate next question is that should we move the operator to the NFD main repo? 😊

yevgeny-shnaidman:
> Thanks @ArangoGutierrez. I think this would make sense.
> 
> With this – the operator basically going hand-in-hand-lockstep with the Helm chart - the immediate next question is that should we move the operator to the NFD main repo? 😊

@marquiz @ArangoGutierrez what is different in this new implementation that requires moving operator into the main repo? If i understand correctly , the new implementation just replaces the build/assets with the helm chats stored in the operator image.

ArangoGutierrez:
> > Thanks @ArangoGutierrez. I think this would make sense.
> > With this – the operator basically going hand-in-hand-lockstep with the Helm chart - the immediate next question is that should we move the operator to the NFD main repo? 😊
> 
> @marquiz @ArangoGutierrez what is different in this new implementation that requires moving operator into the main repo? If i understand correctly , the new implementation just replaces the build/assets with the helm chats stored in the operator image.

Is not a requirement, We have discussed in the past, having the 2 projects in a single git-repo. but is not the goal of this PR. 
What we have discussed in the past, is that the Operator should be the to-go way of deploying NFD, and having it on the same repo makes it more clear for users, and easier to maintain for us. Also Having the Operator be a Helm based operator, means we could potentially have a unified release cycle. instead of the operator playing catch with NFD releases.

config/manager/manager.yaml
```diff
@@ -3,67 +3,95 @@ kind: Namespace
...
+    app.kubernetes.io/part-of: node-feature-discovery-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: system
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: node-feature-discovery-operator
+  namespace: system
```
marquiz, line 18: Is this desired default ns?

marquiz (commented):
Few random'ish comments/questions below.

But I think this is good, if @yevgeny-shnaidman is fine with this change.

Have you verified that it works? 😅

Makefile
```diff
@@ -22,19 +14,19 @@ SITE_BASEURL ?=
...
 # VERSION defines the project version for the bundle.
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
-# - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.2.1)
-# - use environment variables to overwrite this value (e.g export VERSION=0.2.1)
-VERSION := $(shell git describe --tags --dirty --always)
-BUNDLE_VERSION := $(shell git tag | sort -V | tail -1 | awk '{print substr($$1,2); }')
+# - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
+# - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
+VERSION ?= 0.0.1
```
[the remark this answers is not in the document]
ArangoGutierrez: I mean, is the first version of this model , I am open to suggestions

Makefile
```diff
@@ -64,174 +52,203 @@ IMAGE_EXTRA_TAG_NAMES ?=
...
+	mkdir -p $(dir $(HELM_OPERATOR)) ;\
+	curl -sSLo $(HELM_OPERATOR) https://github.com/operator-framework/operator-sdk/releases/download/v1.32.0/helm-operator_$(OS)_$(ARCH) ;\
+	chmod +x $(HELM_OPERATOR) ;\
+	}
+else
+HELM_OPERATOR = $(shell which helm-operator)
+endif
+endif
 
-mdlint:
```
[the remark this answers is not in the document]
ArangoGutierrez: ah ups

config/manager/manager.yaml
ArangoGutierrez: that get's override when deploying, but is good to have it point to where we want from the beginning

Makefile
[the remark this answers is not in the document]
marquiz: What if take it from git (and then can override from the env if needed)

```suggestion
VERSION ?= $(shell git describe --tags --dirty --always)
```

ArangoGutierrez:
> Few random'ish comments/questions below.
> 
> But I think this is good, if @yevgeny-shnaidman is fine with this change.
> 
> Have you verified that it works? 😅

Thanks for the review @marquiz !! 
I have addressed your comments. Please take it for a spin, it works on my local Minikube, but a double check never hurts
PTAL

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -4,30 +4,53 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: node-feature-discovery-operator
+  namespace: node-feature-discovery
```
marquiz, line 7: Argh, sorry about the back-and-forth. But should we actually use the `node-feature-discovery-operator` namespace as default

```suggestion
  namespace: node-feature-discovery-operator
```

Ditto to other operator-related objects

ArangoGutierrez: yup, I was thinking in some NFD stuff while making the change so forgot it was operator 😅

bundle/manifests/nfd-manager-rolebinding_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
```diff
@@ -0,0 +1,20 @@
...
+    app.kubernetes.io/part-of: node-feature-discovery-operator
+  name: nfd-manager-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: nfd-manager-role
+subjects:
+- kind: ServiceAccount
+  name: controller-manager
+  namespace: node-feature-discovery-operator-operator
```
marquiz, line 20: operator-operator?

config/manager/manager.yaml
```diff
@@ -3,67 +3,95 @@ kind: Namespace
...
+    app.kubernetes.io/part-of: node-feature-discovery-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: node-feature-discovery
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: node-feature-discovery-operator
+  namespace: node-feature-discovery-operator-
```
marquiz, line 18: `operator-`?
```suggestion
  namespace: node-feature-discovery-operator
```

marquiz:
Not without issues, did a quick test:

```sh
$ helm -n nfd-o install -f values-local.yaml nfd-o deploy/helm/nfd-operator/
W1103 15:53:54.344149 2245049 warnings.go:70] unknown field "spec.enableNodeFeatureApi"
W1103 15:53:54.344161 2245049 warnings.go:70] unknown field "spec.gc"
...
Error: INSTALLATION FAILED: 1 error occurred:
	* NodeFeatureDiscovery.nfd.k8s-sigs.io "nfd-operator-instance" is invalid: spec.topologyUpdater: Invalid value: "object": spec.topologyUpdater in body must be of type boolean: "object"
```

Also creation of the CR failed:

```sh
$ k apply -f nfd.yaml
Error from server (BadRequest): error when creating "nfd.yaml": NodeFeatureDiscovery in version "v1" cannot be handled as a NodeFeatureDiscovery: strict decoding error: unknown field "spec.image"
```

ArangoGutierrez:
> Not without issues, did a quick test:
> 
> ```shell
> $ helm -n nfd-o install -f values-local.yaml nfd-o deploy/helm/nfd-operator/
> W1103 15:53:54.344149 2245049 warnings.go:70] unknown field "spec.enableNodeFeatureApi"
> W1103 15:53:54.344161 2245049 warnings.go:70] unknown field "spec.gc"
> ...
> Error: INSTALLATION FAILED: 1 error occurred:
> 	* NodeFeatureDiscovery.nfd.k8s-sigs.io "nfd-operator-instance" is invalid: spec.topologyUpdater: Invalid value: "object": spec.topologyUpdater in body must be of type boolean: "object"
> ```
> 
> Also creation of the CR failed:
> 
> ```shell
> $ k apply -f nfd.yaml
> Error from server (BadRequest): error when creating "nfd.yaml": NodeFeatureDiscovery in version "v1" cannot be handled as a NodeFeatureDiscovery: strict decoding error: unknown field "spec.image"
> ```

Thanks for reviewing, I have pushed an amend, I got it working with both `make deploy` and `helm install`

Makefile
```diff
@@ -63,143 +55,200 @@ IMAGE_TAG_NAME ?= $(VERSION)
 IMAGE_EXTRA_TAG_NAMES ?=
 IMAGE_REPO ?= $(IMAGE_REGISTRY)/$(IMAGE_NAME)
 IMAGE_TAG ?= $(IMAGE_REPO):$(IMAGE_TAG_NAME)
-IMAGE_EXTRA_TAGS := $(foreach tag,$(IMAGE_EXTRA_TAG_NAMES),$(IMAGE_REPO):$(tag))
-BUILDER_IMAGE ?= golang:$(GO_VERSION)-bullseye
-BASE_IMAGE_DEBUG ?= debian:buster-slim
-BASE_IMAGE_PROD ?= gcr.io/distroless/base
+IMAGE_EXTRA_TAGS := $(foreach tag,$(IMAGE_EXTRA_TAG_NAMES),$(IMAGE_REPO):$(TAG))
```
marquiz, line 58: This does not look right

config/samples/nodefeaturediscoveries.nfd.k8s-sigs.io.yaml
```diff
@@ -0,0 +1,157 @@
+apiVersion: nfd.k8s-sigs.io/v1
+kind: NodeFeatureDiscovery
+metadata:
+  annotations:
+    helm.sdk.operatorframework.io/uninstall-wait: "true"
+  name: nodefeaturediscovery-instance
+  namespace: node-feature-discovery-operator
```
marquiz, line 7: Hmm, I think here (for the operand) we'd like to use a different namespace(?)
```suggestion
  namespace: node-feature-discovery
```

deploy/helm/nfd-operator/templates/deployment.yaml
```diff
@@ -4,78 +4,54 @@ metadata:
   labels:
     control-plane: controller-manager
     {{- include "node-feature-discovery-operator.labels" . | nindent 4 }}
-  name: {{ include "node-feature-discovery-operator.fullname" . }}-controller-manager
+  name: {{ include "node-feature-discovery-operator.name" . }}-controller-manager
```
marquiz, line 7: Why change this fullname -> name?
```suggestion
  name: {{ include "node-feature-discovery-operator.fullname" . }}-controller-manager
```

deploy/helm/nfd-operator/templates/deployment.yaml
```diff
@@ -4,78 +4,54 @@ metadata:
...
   namespace: {{ include "node-feature-discovery-operator.namespace" . }}
 spec:
-  replicas: 1
   selector:
     matchLabels:
       control-plane: controller-manager
-      {{- include "node-feature-discovery-operator.selectorLabels" . | nindent 6 }}
```
marquiz, line 14: Why remove this?

deploy/helm/nfd-operator/templates/nodefeaturediscoveries.yaml
marquiz, line 1: Do we really want to install the operand as well, by default? And in the same namespace as the operator?

I think also breaks the uninstallation of the operator (helm uninstall will remove the operator but will leave the operand running as the operator does not have time to delete it before being killed itself)

deploy/helm/nfd-operator/templates/rbac.yaml
```diff
@@ -21,67 +27,45 @@ rules:
...
-  - list
-  - watch
-- apiGroups:
-  - apps
-  resources:
-  - deployments
-  - daemonsets
-  - replicasets
-  - statefulsets
+  - secrets
```
marquiz, line 60: Does the operator really need cluster-wide full access to secrets?

deploy/helm/nfd-operator/templates/rbac.yaml
```diff
@@ -1,17 +1,23 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    {{- include "node-feature-discovery-operator.labels" . | nindent 4 }}
+  name: {{ include "node-feature-discovery-operator.name" . }}-service-account
```
marquiz, line 6: Again, shouldn't we use fullname here? Also the `-service-account` feels redundant here

```suggestion
  name: {{ include "node-feature-discovery-operator.fullname" . }}
```

ditto other `name:`s

deploy/helm/nfd-operator/values.yaml
```diff
@@ -3,20 +3,515 @@
...
   # tag, if defined will use the given image tag, else Chart.AppVersion will be used
-  # tag
+  # tag: devel
 imagePullSecrets: []
+labels: {}
+selectorLabels: {}
 
-nameOverride: ""
-fullnameOverride: ""
-namespaceOverride: ""
```
marquiz, line 15: If we delete these, then the templates/_helpers.tpl  should be adjusted accordingly.

These are basically for components that would have nfd-operator as a dependency (so that they would be able to install it in a separate ns etc.) Do we want to preserve this capability? If not, then we can remove these

docs/deployment/helm.md
```diff
@@ -86,14 +86,7 @@ We have introduced the following Chart parameters.
...
 
-| Name | Type | Default | description |
-| ---- | ---- | ------- | ----------- |
-| `image.repository` | string | `{{ site.container_image | split: ":" | first }}` | NFD image repository |
-| `image.tag` | string | `{{ site.release }}` | NFD image tag |
-| `image.pullPolicy` | string | `Always` | Image pull policy |
-| `imagePullSecrets` | list | [] | ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the images used by this PodSpec. If specified, these secrets will be passed to individual puller implementations for them to use. For example, in the case of docker, only DockerConfig type secrets are honored. [More info](https://kubernetes.io/docs/concepts/containers/images#specifying-imagepullsecrets-on-a-pod) |
-| `nameOverride` | string |  | Override the name of the chart |
-| `fullnameOverride` | string |  | Override a default fully qualified app name |
+Check [nfd-chart-params][nfd-chart-params] for a fully detailed list.
```
marquiz, line 89: I think we should still have a table of Helm parameters for the _operator_ itself

marquiz (commented):
Few notes about the deployment. Is it desired that deploying the operator also always automatically deploys the operand in the same namespace? This doesn't feel right to me...

@yevgeny-shnaidman any comments from you to this PR.

ArangoGutierrez:
> Few notes about the deployment. Is it desired that deploying the operator also always automatically deploys the operand in the same namespace? This doesn't feel right to me...
> 
> @yevgeny-shnaidman any comments from you to this PR.

Umm It was an idea I had, to help users get a running NFD right after `helm install` but can be removed.

config/manager/kustomization.yaml
```diff
@@ -1,16 +1,8 @@
...
-  - controller_manager_config.yaml
-  name: manager-config
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-nfd/node-feature-discovery-operator
-  newTag: master
+  newName: ghcr.io/arangogutierrez/node-feature-discovery-operator
+  newTag: v0.2.0-226-g4008131c-dirty
```
marquiz, line 8: ??

deploy/helm/nfd-operator/templates/deployment.yaml
```diff
@@ -7,75 +7,52 @@ metadata:
...
     matchLabels:
       control-plane: controller-manager
       {{- include "node-feature-discovery-operator.selectorLabels" . | nindent 6 }}
+  replicas: 1
   template:
     metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
       labels:
-        {{- include "node-feature-discovery-operator.selectorLabels" . | nindent 8 }}
```
marquiz, line 18: You must not delete the selectorLabels from here as you have them in the selector above

marquiz:
> Umm It was an idea I had, to help users get a running NFD right after `helm install` but can be removed.

Kinda nice, I guess, but if we do this we should have a clear understanding of the implications. In this scenario the operator and operand become part of the same deployment (also the operand will be managed through the same helm deployment) 🧐 At least we should have a parameter for the user to opt-out from default-installing the operand. Plus make sure that helm uninstall works as expected in all cases (which is hard, I think, as the Helm deployment is not managing the nfd operand itself...)

config/manager/kustomization.yaml
ArangoGutierrez: Kustomize.... grrr my bad, forgot to restore here

yevgeny-shnaidman:
> Few notes about the deployment. Is it desired that deploying the operator also always automatically deploys the operand in the same namespace? This doesn't feel right to me...
> 
> @yevgeny-shnaidman any comments from you to this PR.

@marquiz  if i understand correctly than the operand will only be deployed once the NFD CR is deployed. or does it bother you that it will be deployed in the same namespace?

marquiz:
> @marquiz if i understand correctly than the operand will only be deployed once the NFD CR is deployed. or does it bother you that it will be deployed in the same namespace?

Probably the biggest concern were the dependencies in the deployment and who manages what.

I.e. NFD CR was deployed by Helm -> the operator deployed NFD operand -> Helm uninstall will remove NFD CR and the operator and the NFD remains running un-managed

I had hard time wrapping my head around this scheme. At least there should be a parameter to disable automatic deployment of NFD CR (and thus the operand). Also it feels a bit overloaded/unorthodox to manage both the operator and the operator with the same Helm chart (what is the purpose of the operator in this scheme?)

yevgeny-shnaidman:
> > @marquiz if i understand correctly than the operand will only be deployed once the NFD CR is deployed. or does it bother you that it will be deployed in the same namespace?
> 
> Probably the biggest concern were the dependencies in the deployment and who manages what.
> 
> I.e. NFD CR was deployed by Helm -> the operator deployed NFD operand -> Helm uninstall will remove NFD CR and the operator and the NFD remains running un-managed
> 
> I had hard time wrapping my head around this scheme. At least there should be a parameter to disable automatic deployment of NFD CR (and thus the operand). Also it feels a bit overloaded/unorthodox to manage both the operator and the operator with the same Helm chart (what is the purpose of the operator in this scheme?)

I am probably missing something, what do you mean by automatic deployment of NFD CR? NFD CR is deployed by cluster admin/user/other operator, no?  If i understand the sequence correctly:
1. operator is deployed
2. admin/user/other_operator deploys NFD CR ( either via Helm or by some other means)
3. operator deploys operand
In case operator is un-deployed at the same time as NFD CR is un-deployed, the it is a problem, but maybe we can fix it with Owner/Controller references

deploy/helm/nfd-operator/templates/rbac.yaml
```diff
@@ -350,20 +363,5 @@ roleRef:
...
-    {{- include "node-feature-discovery-operator.labels" . | nindent 4 }}
-  name: {{ include "node-feature-discovery-operator.fullname" . }}-proxy
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: {{ include "node-feature-discovery-operator.fullname" . }}-proxy
-subjects:
-- kind: ServiceAccount
-  name: default
+  name: {{ include "node-feature-discovery-operator.name" . }}-service-account
```
marquiz, line 366:
```suggestion
  name: {{ include "node-feature-discovery-operator.fullname" . }}
```

marquiz (commented):
Looks like we need the clusterrole to be able to access secrets for the Helm operator to be able to work. My bad

marquiz:
> NFD CR is deployed by cluster admin/user/other operator, no? If i understand the sequence correctly:
> 
> 1. operator is deployed
> 2. admin/user/other_operator deploys NFD CR ( either via Helm or by some other means)
> 3. operator deploys operand
>    In case operator is un-deployed at the same time as NFD CR is un-deployed, the it is a problem, but maybe we can fix it with Owner/Controller references

Exactly my thinking. I was commenting a previous version of this PR where step 2. was missing (or steps 1. and 2. were combined). But now that's gone from the PR

ArangoGutierrez:
> > NFD CR is deployed by cluster admin/user/other operator, no? If i understand the sequence correctly:
> > 
> > 1. operator is deployed
> > 2. admin/user/other_operator deploys NFD CR ( either via Helm or by some other means)
> > 3. operator deploys operand
> >    In case operator is un-deployed at the same time as NFD CR is un-deployed, the it is a problem, but maybe we can fix it with Owner/Controller references
> 
> Exactly my thinking. I was commenting a previous version of this PR where step 2. was missing (or steps 1. and 2. were combined). But now that's gone from the PR

I think we can add the automated CR deployment later on, just moving to helm is a big change, we don't want to change today's way of deploying NFD, people are used to deploying the CR after deploying via Helm/Kustomize, let's not surprise them with a pre-created CR for now

deploy/helm/nfd-operator/templates/rbac.yaml
```diff
@@ -151,7 +139,7 @@ rules:
   - get
   - list
   - patch
-  - update
+  - updates
```
marquiz, line 142: updates??
```suggestion
  - update
```

ArangoGutierrez:
@yevgeny-shnaidman have you had time to take this Patch for a spin?

docs/deployment/helm.md
```diff
@@ -102,4 +117,6 @@ We have introduced the following Chart parameters.
 | `controller.image.repository` | string | `{{ site.container_image | split: ":" | first }}` | NFD-Operator image repository |
 | `controller.image.tag` | string | `{{ site.release }}` | NFD-Operator image tag |
 
+<!-- Links -->
 [rbac]: https://kubernetes.io/docs/reference/access-authn-authz/rbac/
+[nfd-chart-params]: https://kubernetes-sigs.github.io/node-feature-discovery/v0.14/deployment/helm.html#chart-parameters
```
marquiz, line 122: Stale link
```suggestion
```

docs/deployment/helm.md
```diff
@@ -44,26 +44,39 @@ repository and install from there.
...
 helm install nfd-operator ./nfd-operator/ --namespace $NFD_O_NS --create-namespace
 ```
 
 See the [configuration](#configuration) section below for instructions how to
 alter the deployment parameters.
 
+By default the operator will watch `NodeFeatureDiscovery` objects
+only in the namespace where the operator is deployed in. This is
+specified by the `WATCH_NAMESPACE` env variable in the operator
+deployment manifest. If unset the operator will watch ALL
```
marquiz, line 57: I don't see the operator behaving this way 🧐 Should it?

ArangoGutierrez: ah ups, that's old school behav

yevgeny-shnaidman:
> @yevgeny-shnaidman have you had time to take this Patch for a spin?

i am trying to port into OCP as we speak...

yevgeny-shnaidman:
> > @yevgeny-shnaidman have you had time to take this Patch for a spin?
> 
> i am trying to port into OCP as we speak...

BTW, is there a helm-operator image for ARM?

ArangoGutierrez:
> > > @yevgeny-shnaidman have you had time to take this Patch for a spin?
> > 
> > 
> > i am trying to port into OCP as we speak...
> 
> BTW, is there a helm-operator image for ARM?

I built and tested this on my M1 Mac, just run `make image` and you will get an ARM image.

yevgeny-shnaidman:
> > > > @yevgeny-shnaidman have you had time to take this Patch for a spin?
> > > 
> > > 
> > > i am trying to port into OCP as we speak...
> > 
> > 
> > BTW, is there a helm-operator image for ARM?
> 
> I built and tested this on my M1 Mac, just run `make image` and you will get an ARM image.

It' s a multi-arch image?

ArangoGutierrez:
> > > > > @yevgeny-shnaidman have you had time to take this Patch for a spin?
> > > > 
> > > > 
> > > > i am trying to port into OCP as we speak...
> > > 
> > > 
> > > BTW, is there a helm-operator image for ARM?
> > 
> > 
> > I built and tested this on my M1 Mac, just run `make image` and you will get an ARM image.
> 
> It' s a multi-arch image?

you can build multiarch if you want, look at the Makefile rules, I think you need docker with buildx enabled in your host

nfd/node-feature-discovery/templates/topologyupdater.yaml
```diff
@@ -0,0 +1,156 @@
+{{- if .Values.topologyUpdater.enable -}}
```
yevgeny-shnaidman, line 1: we want those values to be taken from NFD CR, no? in that case it should be:

{{- if .Values.topologyUpdater -}}

nfd/node-feature-discovery/templates/topologyupdater.yaml
```diff
@@ -0,0 +1,156 @@
...
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "node-feature-discovery.fullname" . }}-topology-updater
```
yevgeny-shnaidman, line 5: General question for all the names in the templates:  currently ( before this PR) all the names are hardcoded. Are we expecting to support more then one NFD CR per cluster?

nfd/node-feature-discovery/templates/worker.yaml
```diff
@@ -0,0 +1,152 @@
...
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      serviceAccountName: {{ include "node-feature-discovery.worker.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.worker.podSecurityContext | nindent 8 }}
+      containers:
+      - name: worker
+        securityContext:
+          {{- toYaml .Values.worker.securityContext | nindent 12 }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
yevgeny-shnaidman, line 40: same here: don't we want to take the image from the NFD CR?

ArangoGutierrez:
@yevgeny-shnaidman  Everything under the `nfd` folder has to be the same Helm chart as in the NFD Helm chart. we should not modify that folder.

yevgeny-shnaidman:
> @yevgeny-shnaidman Everything under the `nfd` folder has to be the same Helm chart as in the NFD Helm chart. we should not modify that folder.
> 
> The `helm-based-operator` will use the information on the CR to then deploy the chart under `nfd` folder and modify the variables as needed.

@ArangoGutierrez i understand that, i am just saying that with current code, the image values ( for example) are not taken from the NFD CR

yevgeny-shnaidman:
> > > @yevgeny-shnaidman Everything under the `nfd` folder has to be the same Helm chart as in the NFD Helm chart. we should not modify that folder.
> > > The `helm-based-operator` will use the information on the CR to then deploy the chart under `nfd` folder and modify the variables as needed.
> > 
> > 
> > @ArangoGutierrez i understand that, i am just saying that with current code, the image values ( for example) are not taken from the NFD CR
> 
> Why you say that? the NFD folder got created by the `operator-framework` tool, and it ties the Helm Values.yaml file to the CRD. remeber this PR creates a new CRD that is basically NFD Values.yaml file

Right, i have missed that part ( new CRD). I have been using the existing CR in my test. But do we want to change the CRD?  Can't we use the old one?  New CRD is a breaking change..

ArangoGutierrez:
> > > > @yevgeny-shnaidman Everything under the `nfd` folder has to be the same Helm chart as in the NFD Helm chart. we should not modify that folder.
> 
> > > > The `helm-based-operator` will use the information on the CR to then deploy the chart under `nfd` folder and modify the variables as needed.
> 
> > > 
> 
> > > 
> 
> > > @ArangoGutierrez i understand that, i am just saying that with current code, the image values ( for example) are not taken from the NFD CR
> 
> > 
> 
> > Why you say that? the NFD folder got created by the `operator-framework` tool, and it ties the Helm Values.yaml file to the CRD. remeber this PR creates a new CRD that is basically NFD Values.yaml file
> 
> 
> 
> Right, i have missed that part ( new CRD). I have been using the existing CR in my test. But do we want to change the CRD?  Can't we use the old one?  New CRD is a breaking change..

Yes, we know is a hard change, but is an improvement for the better of the project, and allows better integration with NFD deployment

ArangoGutierrez:
Hi @yevgeny-shnaidman , I know this is a big change but we will get a lot from it. Migrating won't be that hard, we could help with documentation and examples.
Hope we can work together to build a better Operator that goes more tied coupled with the Operand it manages

yevgeny-shnaidman:
> Hi @yevgeny-shnaidman , I know this is a big change but we will get a lot from it. Migrating won't be that hard, we could help with documentation and examples. Hope we can work together to build a better Operator that goes more tied coupled with the Operand it manages

@ArangoGutierrez i think think that if we keep the NFD CRD as is we can integrate your PR pretty easily.  All we need to do is change the values.yaml to be conformant with the NFD CRD, and to change templates' conditions  accordingly. Any new configuration (GC, Topology, etc') can be added as new fields to the CRD, so that most customer can continue using their operators without any change.

ArangoGutierrez:
> > Hi @yevgeny-shnaidman , I know this is a big change but we will get a lot from it. Migrating won't be that hard, we could help with documentation and examples. Hope we can work together to build a better Operator that goes more tied coupled with the Operand it manages
> 
> @ArangoGutierrez i think think that if we keep the NFD CRD as is we can integrate your PR pretty easily. All we need to do is change the values.yaml to be conformant with the NFD CRD, and to change templates' conditions accordingly. Any new configuration (GC, Topology, etc') can be added as new fields to the CRD, so that most customer can continue using their operators without any change.

Changing the `values.yaml` is not an option, the idea is to use the one straight from NFD Charts, if not we end up in the same situation where the operator has a "special" condition that we have to maintain, and is extra work. The whole idea behind this PR, and other comments like moving the Operator to the NFD repo, is to reduce the maintenance effort, so we can focus on features. 

We understand this might be a big change, but moving from the current CRD to this new one is not a lot of effort. and right now the operator is so outdated, that customers can not use new features from NFD, and that is something we want to remediate.

PR #211, opened 2024-02-12
controller-runtime package update

yevgeny-shnaidman:
updating controller package to v0.17.0.
updating the main, in the manager initialization code to support the v0.17.0


main.go
```diff
@@ -91,13 +94,22 @@ func main() {
...
 		HealthProbeBindAddress: args.probeAddr,
 		LeaderElection:         args.enableLeaderElection,
 		LeaderElectionID:       "39f5e5c3.nodefeaturediscoveries.nfd.kubernetes.io",
-		Namespace:              watchNamespace,
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				watchNamespace: cache.Config{},
+			},
+		},
+		//Namespace:              watchNamespace,
```
marquiz, line 112: Why is this commented out?

marquiz (commented):
One comment below. Otherwise looks good to me

mythi:
> updating controller package to v0.17.0.

`v0.17.1` was released on Friday and comes with a bug fix so perhaps update to that directly.

main.go
yevgeny-shnaidman: Sorry, my bad , forgot to remove the comment. with v0.17.0 the manager Option do not have the "Namespace" field anymore, it have been moved into the Cache field

main.go
```diff
@@ -91,13 +94,21 @@ func main() {
...
+		WebhookServer: webhook.NewServer(webhook.Options{
+			Port: 9443,
+		}),
 		HealthProbeBindAddress: args.probeAddr,
 		LeaderElection:         args.enableLeaderElection,
 		LeaderElectionID:       "39f5e5c3.nodefeaturediscoveries.nfd.kubernetes.io",
-		Namespace:              watchNamespace,
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				watchNamespace: cache.Config{},
```
ArangoGutierrez, line 109: We have a `utils.GetWatchNamespace()` is this no longer needed?

yevgeny-shnaidman: it is needed. The Namespace parameter of the manager Options in 0.14.0 has been moved into the cache.Options. DefaultNamespaces is a map that can configure all the namespaces to watch  (for cache). We are using the output value of utils.GetWatchNamespace()  as a key in the DefaultNamespaces map

marquiz (approved):
Thankls @yevgeny-shnaidman for the patch. FWIW, +1 from me 😇

marquiz:
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #212, opened 2024-02-13
Adding new controller template

yevgeny-shnaidman:
This PR is part of the refactoring effort. It contains the following: 1) new controller template implementation. It contains only the sequence
   of initialization for various components (master, worker, topology
etc'). The actual implementation will be done via implementing the helper interface functions
2) unit-test for the controller
3) adding unit-test target to the Makefile
4) adding mockgen and golangci-lint downloading targets to the Makefile
   mockgen is used for generating the mock files for the interface.


ArangoGutierrez:
Wow awesome! will take a look soon

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -0,0 +1,178 @@
+/*
+Copyright 2020-2021 The Kubernetes Authors.
```
marquiz, line 2:
```suggestion
Copyright 2024 The Kubernetes Authors.
```

Also check that all the newly added (non-autogenerated) files have the same license header as this

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -0,0 +1,178 @@
...
+		GenericFunc: func(event.GenericEvent) bool { return false },
+	}
+}
+
+// +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=nfd.k8s-sigs.io,resources=nodefeaturerules,verbs=get;list;watch
+
+// to move the current state of the cluster closer to the desired state.
```
marquiz, line 75: nit: the comment should start with "Reconcile", e.g.  `Reconcile moves the ...`

internal/controllers/suite_test.go
```diff
@@ -0,0 +1,40 @@
...
+package new_controllers
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/node-feature-discovery-operator/internal/test"
+	//+kubebuilder:scaffold:imports
```
marquiz, line 26: Just a question to educate myself: what is this doing? 😇 Can you point some documentation?

marquiz (commented):
Thanks @yevgeny-shnaidman for the PR. Just a few small notes from me.

To make this PR cleaner and leaner could we split out the changes to the Makefile in separate PR(s)?

yevgeny-shnaidman:
> Thanks @yevgeny-shnaidman for the PR. Just a few small notes from me.
> 
> To make this PR cleaner and leaner could we split out the changes to the Makefile in separate PR(s)?

@marquiz thank you for the review, moved the Makefile changes to a different [PR](https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/213)

internal/controllers/nodefeaturediscovery_reconciler.go
yevgeny-shnaidman: Fixed

internal/controllers/suite_test.go
yevgeny-shnaidman: This marker is set by kubebuilder during project bootstrap.  So it actually appears only in file that are created during bootstrap, but is missing from the files that are created later by developers. It's purpose is to manage the import of the Go file ( go over the code and adding the imports as needed), but i have never seen it used, since developers usually do the imports as part of developing code.  I can remove it, if you think it clutters the code

marquiz: Your call, I think we can keep it if/when it's "standard clutter". I only wanted to understand better as this is the first time seeing this marker

yevgeny-shnaidman: ok, than i will keep for now, unless we get additional complaints

internal/controllers/suite_test.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2024.
```
marquiz, line 2: nit: use the standard one

```suggestion
Copyright 2024 The Kubernetes Authors.
```

marquiz (approved):
Thanks @yevgeny-shnaidman. For an empty scaffolding, looks good to me.
/assign @ArangoGutierrez

ArangoGutierrez:
> > Thanks @yevgeny-shnaidman for the PR. Just a few small notes from me.
> > To make this PR cleaner and leaner could we split out the changes to the Makefile in separate PR(s)?
> 
> @marquiz thank you for the review, moved the Makefile changes to a different [PR](https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/213)

/lgtm

PR #213, opened 2024-02-14
adding targets to Malefile

yevgeny-shnaidman:
1) installing mockgen to support mocks generation
   for unit test
2) generate target will also generate the mock files using mockgen 3) ading unit-test target to run unit-test
4) adding golangci-lint target to install golangci-lint under
   PROJECT/bin. Currently it will be a standalone, that will allow to
manually run verify target prior to pushing commit


Makefile
```diff
@@ -143,6 +143,12 @@ fmt:
 vet:
 	$(GO_CMD)  vet ./...
 
+TEST ?= ./...
```
marquiz, line 146: Just a thought/suggestion, not a blocker. I think a very common pattern is to call this kind of variable `WHAT`. Also, then it could be used in some other targets like `vet:` above.

```suggestion
WHAT ?= ./...
```

WDYT? Or alternatively suggestion to change the name slightly:

```suggestion
TESTS ?= ./...
```

marquiz (commented):
Thanks @yevgeny-shnaidman for splitting this into a separate PR. Easier to review this way.

Makefile
```diff
@@ -213,6 +220,15 @@ CONTROLLER_GEN = $(PROJECT_DIR)/bin/controller-gen
 controller-gen:
 	@GOBIN=$(PROJECT_DIR)/bin GO111MODULE=on $(GO_CMD) install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
 
+.PHONY: mockgen
+mockgen: ## Install mockgen locally.
+	$(GO_CMD) install go.uber.org/mock/mockgen@v0.3.0
+
+GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
+.PHONY: golangci-lint
+golangci-lint: ## Download golangci-lint locally if necessary.
```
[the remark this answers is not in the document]
yevgeny-shnaidman: it just runs the install

Makefile
yevgeny-shnaidman: I would change it to TESTS. So far from my experience (not that extensive :), i have seen the TEST/TESTS being used , but have yet to see WHAT

yevgeny-shnaidman: fixed

Makefile
```diff
@@ -213,6 +220,15 @@ CONTROLLER_GEN = $(PROJECT_DIR)/bin/controller-gen
...
+	@GOBIN=$(PROJECT_DIR)/bin  GO111MODULE=on $(GO_CMD) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed to 1.56.1

PR #215, opened 2024-02-19
skipperizing the repo

yevgeny-shnaidman:
skipper is a tool that allows executing all Makefile commands in the container enviroment, without any need to install project dependencies on the host. It is especially useful when working on different multiple projects. All it requires is installing docker and skipper on the host.
skipper is used besides the Makefile, meaning: if somebody does not want to use skipper, he can run makefile commands as usual. skipper can be used on Golang project, Python projects and almost any other programical language
Example:
1) without skipper:
    install GO on your host
    run: make build
         make image
         make deploy
2) with skipper:
   run:
        skipper make build
        skipper make make
        skipper make deploy

skipper also supports passing environment variable to Makefile. skipper repo: https://github.com/Stratoscale/skipper


skipper.yaml
```diff
@@ -0,0 +1,16 @@
+registry: quay.io
```
marquiz, line 1: Does this matter (is it needed). Just `quay.io` looks a bit random 🤔 Could it be just empty string, or `local` or `registry.k8s.io` WDYT?

skipper.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+containers:
+  node-feature-discovery-operator-build: Dockerfile.node-feature-discovery-operator-build
+volumes:
+    - $HOME/.cache/go-build:$HOME/.cache/go-build
+    - $HOME/go/pkg:/go/pkg
+    - $HOME/.cache/golangci-lint:$HOME/.cache/golangci-lint
+    - $HOME/.docker:$HOME/.docker
+    - $HOME/.kube/:$HOME/.kube
+    - $HOME/.minikube/:$HOME/.minikube
```
marquiz, line 12: Are these needed for something? Just thinking the principle of least privileges/smallest blast radius

marquiz (commented):
Thanks @yevgeny-shnaidman for the contribution.

Could we also have a small update on the documentation (`docs/advanced/developer-guide.md`). For example, describe the same things as in the commit message

skipper.yaml
yevgeny-shnaidman: the goal of "registry" is to define the registry where the skipper-built image can be pushed. But in practice, nobody ever pushes it, people just build the image once, locally, if needed, and then use it. I will change it to local

skipper.yaml
yevgeny-shnaidman: i will remove kube and minikube mappings.If i am not mistaken, then docker mappings are needed

yevgeny-shnaidman:
> Thanks @yevgeny-shnaidman for the contribution.
> 
> Could we also have a small update on the documentation (`docs/advanced/developer-guide.md`). For example, describe the same things as in the commit message

@marquiz added a skipper section to the docs

marquiz (approved):
Nice improvement @yevgeny-shnaidman

marquiz:
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #217, opened 2024-03-03
Adding topology deployment implementation

yevgeny-shnaidman:
On every reconciliation loop, the controller updates/creates topology daemonset if needed, based on the NFD CR spec.


internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,174 @@
...
+func getImagePullPolicy(nfdInstance *nfdv1.NodeFeatureDiscovery) corev1.PullPolicy {
+	if nfdInstance.Spec.Operand.ImagePullPolicy != "" {
+		return corev1.PullPolicy(nfdInstance.Spec.Operand.ImagePullPolicy)
+	}
+	return corev1.PullAlways
+}
+
+func getArgs(nfdInstance *nfdv1.NodeFeatureDiscovery) []string {
+	return []string{
+		"-kubelet-config-uri=file:/host-var/lib/kubelet/config.yaml",
```
marquiz, line 93: Do we really need this (and the mounts)? NFD uses the configz endpoint by default.

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+}
+
+func getVolumes() []corev1.Volume {
+	return []corev1.Volume{
+		{
+			Name: "kubelet-podresources-conf",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/etc/kubernetes/kubelet.conf",
```
marquiz, line 150: IF we want to use the config file (instead of configz which is the default), I think we should mount `/var/lib/kubelet/config.yaml`. Isn't `/etc/kubernetes/kubelet.conf` a kubeconfig file (for kubelet) which we are not interested in(?)

internal/daemonset/daemonset.go
```diff
@@ -0,0 +1,174 @@
...
+		},
+		{
+			Name: "host-sys",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/sys",
+					Type: ptr.To[corev1.HostPathType](corev1.HostPathDirectory),
+				},
+			},
+		},
```
marquiz, line 172: I think we'd want to add the volume (and mount) for `-kubelet-state-dir` i.e. `/var/lib/kubelet` from the host

marquiz (commented):
Thanks @yevgeny-shnaidman. Sorry for the silence, I was on PTO for over a week. Some comments about mounts/volumes but otherwise looks good.

/assign @PiotrProkop

internal/daemonset/daemonset.go
yevgeny-shnaidman: currently we will go with the bare minimum, if the configz defined in the main as default, we won't provide  the option to reconfigure. This can all be done at the later point, when  a need arises

internal/daemonset/daemonset.go
yevgeny-shnaidman: ok , sure, removed the argument and the mount.

marquiz (approved):
Thanks @yevgeny-shnaidman for the update. Looks good, I haven't tested it though 😅

@ArangoGutierrez @PiotrProkop PTAL

yevgeny-shnaidman:
> Thanks @yevgeny-shnaidman for the update. Looks good, I haven't tested it though 😅
> 
> @ArangoGutierrez @PiotrProkop PTAL

@marquiz curently we are just adding the building blocks. Once we have coded everything (master, worker, topology, prune , gc) we will add configs and then we can start really testing. I am pretty sure we will add more changes at that stage :)

Issue #210, opened 2024-01-14
Re-factor operator code to be compliant with k8s operator guidelines


yevgeny-shnaidman:
Currently the NFD operator is implemented using the outdated operator format, relaying on the hard-coded
yamls files, configured in the build/assets directory.  Those directories include, besides Deployments and Daemonsets definitions also service-accounts/roles/roles/bindings definitions.Reconciliation loop is reading all those files, updates the data with input from NFD CR and creates/updates the resources in the k8s cluster.

We want to  move to a more compliant operator, and those are the changes that we may apply:
1.Remove build/assets directory. All the k8s objects will be created on-the-fly by the appropriate reconciller/s with the default values of the structures updated with values taken from NFD CR
2. All the RBAC/SA  object will not be created/updated during reconciliation, but will be created during the deployment process
3. Reconcile the following 3 components that can be deployed by the operator: 
      - worker daemonset/configmap
      - master deployment
      - topology daemonset/configmap

We have 2 option how to reconcile those objects:
### Option 1 ###
 We define a controller per component (worker, master, topology). Each controller will watch for NFD CR and the appropriate 
 k8s object (daemonset, deployment, configmap) and will run a reconciliation loop whenever it detects a change. 
### Option 2 ###
We define one controller that will watch for NFD CR, daemonsets/deployments and configmaps. It will run a reconciliation on 
any change. It will call a dedicated package to reconcile each one of the components. 

### CR Status ###
We can remove removing reporting conditions on any of the RBAC features/SA, since those will be created during operator deployment. We will only report on Daemonsets/Deployment creation/patching. No need to report on configmaps

marquiz:
Thanks @yevgeny-shnaidman for this. The plan sounds good to me 👍 I would've probably resorted to e.g. ansible-based operator had I been doing this overhaul myself 😅 But you know operators and e.g. the OpenShift requirements so much better than me (still an operator n00b).

Regarding option 1 vs option 2: are you still talking about one NFD CRD (not one CRD per daemon)? I don't think I have much opinion on which would be preferred. Whatever makes sense from the implementation pov, I'd choose...

Thoughts @ArangoGutierrez?

yevgeny-shnaidman:
@marquiz we are not changing/adding any APIs (CRDs) , just refactoring the code. a controller or controllers will watch  for NFD CR and will extract from it the parts that are relevant for their specific components

yevgeny-shnaidman:
[...]

PR #218, opened 2024-03-11
Add worker daemonsets to refactored reconcile loop

chr15p:
On every reconciliation loop the controller will check/create/update the worker daemonset and configmap if needed, based on the NFD CR spec. This forms part of the Operator Refactoring work ( #210 )

This PR also sets up the  role/rolebindings etc needed for this worker to be created by kustomize at operator install time.

Notes:
- All of the controller logic is in the `internal/worker/` directory because it is dealing with a daemonset and a configmap, so `internal/daemonset/` doesn't make a lot of sense to me.
- The configmap draws its contents from the NodeFeatureDiscovery resource because it makes sense (to me at least) to make any changes in one place (the NFD cr) rather than risk confusion with what gets changed where.
- The permissions are being set in the `config/worker/` directory, again to try and keep everything relevant to the worker in a single place, they are copied form the `build/worker`  so should be entirely compatible.

other opinions on those are of course welcome.


config/default/kustomization.yaml
```diff
@@ -24,3 +24,4 @@ resources:
 - ../crd
 - ../rbac
```
yevgeny-shnaidman, line 25: @chr15p lets keep the config changes for later PRs. We would like to get all the controllers/pkg code ready, and once it is done we will add the configuration and switch to the new code. This way we are keeping PRs smaller and also let the operator be deploy-able (with the old version) at the same time

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -177,6 +181,34 @@ func (nfdh *nodeFeatureDiscoveryHelper) handleMaster(ctx context.Context, nfdIns
...
+	workerCM := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: "nfd-worker", Namespace: nfdInstance.Namespace},
+	}
+	cmRes, err := controllerutil.CreateOrPatch(ctx, nfdh.client, &workerCM, func() error {
+		return nfdh.workerAPI.SetWorkerConfigMapAsDesired(ctx, nfdInstance, &workerCM)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to reconcile worker configmap %s/%s: %w", nfdInstance.Namespace, nfdInstance.Name, err)
+	}
+    logger.Info("reconciled worker DaemonSet", "namespace", nfdInstance.Namespace, "name", nfdInstance.Name, "result", cmRes)
```
yevgeny-shnaidman, line 195:
```suggestion
    logger.Info("reconciled worker ConfigMap", "namespace", nfdInstance.Namespace, "name", nfdInstance.Name, "result", cmRes)
```

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -42,8 +43,8 @@ type nodeFeatureDiscoveryReconciler struct {
 	helper nodeFeatureDiscoveryHelperAPI
 }
 
-func NewNodeFeatureDiscoveryReconciler(client client.Client, deploymentAPI deployment.DeploymentAPI, scheme *runtime.Scheme) nodeFeatureDiscoveryReconciler {
-	helper := newNodeFeatureDiscoveryHelperAPI(client, deploymentAPI, scheme)
+func NewNodeFeatureDiscoveryReconciler(client client.Client, deploymentAPI deployment.DeploymentAPI, workerAPI worker.WorkerAPI, scheme *runtime.Scheme) nodeFeatureDiscoveryReconciler {
```
yevgeny-shnaidman, line 46: IMHO it is better to use daemonset/deployment/configmap packages than worker/topology/etc'. Please see my [PR](https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/217).  We daemonsets/configmaps/deployments are the building block: we have daemonsets both for topology and workers, so i think it makes sense to set all daemonset under specific package etc'. Also, in case any other  future component will need to be deployed by the operator, we can just use the building blocks as is, with minimal changes. WDYT?

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -177,6 +181,34 @@ func (nfdh *nodeFeatureDiscoveryHelper) handleMaster(ctx context.Context, nfdIns
...
+	opRes, err := controllerutil.CreateOrPatch(ctx, nfdh.client, &workerDS, func() error {
+		return nfdh.workerAPI.SetWorkerDaemonsetAsDesired(ctx, nfdInstance, &workerDS)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to reconcile worker DaemonSet %s/%s: %w", nfdInstance.Namespace, nfdInstance.Name, err)
+	}
+    // the worker DS now exists
+
+
+    logger.Info("reconciled worker DaemonSet", "namespace", nfdInstance.Namespace, "name", nfdInstance.Name, "result", opRes)
```
yevgeny-shnaidman, line 210: you probably need to run fmt prior to submitting

internal/controllers/nodefeaturediscovery_reconciler.go
marquiz: Sounds sensible to me (if re-usable building blocks like "daemonset" can be easily made)

internal/daemonset/worker.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+	}
+	return containerVolume
+}
+
+func getWorkerLabelsAForApp(name string) map[string]string {
+	return map[string]string{"app": name}
+}
+
+func (d *daemonset) SetWorkerDaemonsetAsDesired(ctx context.Context, nfdInstance *nfdv1.NodeFeatureDiscovery, workerDS *appsv1.DaemonSet) error {
```
yevgeny-shnaidman, line 200: it seems a little strange that SetTopology is defined in daemonset.go and SetWorker is defined in the worker.go. I would suggest that both those function will be defined in the daemonset.go.  The helper functions (getLabels,getTolerations etc') can be defined in a dedicated file (worker.go). In addition we can keep the unit tests defined in the daemonset_test.go, without defining dedicated test file per each function

config/default/kustomization.yaml
```diff
@@ -24,3 +24,4 @@ resources:
...
 - ../manager
+- ../worker
```
yevgeny-shnaidman, line 27: This line is also not needed in this PR

internal/daemonset/worker.go
```diff
@@ -0,0 +1,172 @@
...
+
+package daemonset
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/utils/ptr"
+)
+
+//go:generate mockgen -source=worker.go -package=daemonset -destination=mock_worker.go DaemonsetAPI
```
yevgeny-shnaidman, line 25: no need for this line, there is no interface definition here

yevgeny-shnaidman:
ping @marquiz 
ping @ArangoGutierrez

internal/configmap/testdata/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
```diff
@@ -0,0 +1,123 @@
...
+  namespace: test-namespace
+spec:
+  instance: "" # instance is empty by default
+  #labelWhiteList: ""
+  #extraLabelNs:
+  #  - "example.com"
+  #resourceLabels:
+  #  - "example.com/resource"
+  operand:
+    image: gcr.io/k8s-staging-nfd/node-feature-discovery:master-minimal
```
marquiz, line 14: minor nit: we can just use the `master` tag, it's the same as `master-minimal`

```suggestion
    image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
```

But maybe we can change this with possible other instances in a separate PR.

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -182,6 +185,31 @@ func (nfdh *nodeFeatureDiscoveryHelper) handleMaster(ctx context.Context, nfdIns
...
+	workerCM := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Name: "nfd-worker", Namespace: nfdInstance.Namespace},
+	}
+	cmRes, err := controllerutil.CreateOrPatch(ctx, nfdh.client, &workerCM, func() error {
+		return nfdh.configmapAPI.SetWorkerConfigMapAsDesired(ctx, nfdInstance, &workerCM)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to reconcile worker configmap %s/%s: %w", nfdInstance.Namespace, nfdInstance.Name, err)
+	}
+	logger.Info("reconciled worker DaemonSet", "namespace", nfdInstance.Namespace, "name", nfdInstance.Name, "result", cmRes)
```
marquiz, line 199: configmap?
```suggestion
	logger.Info("reconciled worker ConfigMap", "namespace", nfdInstance.Namespace, "name", nfdInstance.Name, "result", cmRes)
```

internal/daemonset/daemonset.go
```diff
@@ -171,3 +172,44 @@ func getVolumes() []corev1.Volume {
...
+
+				ServiceAccountName: "nfd-worker",
+				DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
+				Containers: []corev1.Container{
+					{
+						Env:             getEnvs(),
+						Image:           nfdInstance.Spec.Operand.ImagePath(),
+						Name:            "nfd-worker",
+						Command:         []string{"nfd-worker"},
+						Args:            []string{"--server=nfd-master:$(NFD_MASTER_SERVICE_PORT)"},
```
marquiz, line 204: Just a note that this is only relevant if the NodeFeature API is disabled and gRPC is enabled (`--enable-nodefeature-api=false` in NFD v0.0.14 and v0.15.0). We could remove this/these especially if we're not planning to have a setting for enabling the (deprecated) gRPC API. WDYT?

```suggestion
						Args:            []string{},
```

It shouldn't hurt to have it, though (just ignored by default as NodeFeature API is enabled)

internal/daemonset/worker.go
```diff
@@ -0,0 +1,170 @@
...
+func getWorkerVolumes() []corev1.Volume {
+	containerVolume := []corev1.Volume{
+		{
+			Name: "host-dev",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/dev",
+				},
+			},
+		},
```
marquiz, line 109:
```suggestion
```

internal/daemonset/worker.go
```diff
@@ -0,0 +1,170 @@
...
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		},
+	}
+}
+
+func getWorkerVolumeMounts() *[]corev1.VolumeMount {
+
+	containerVolumeMounts := []corev1.VolumeMount{
+		{
```
marquiz, line 69: In the upstream NFD Helm chart (upcoming v0.16) we have host mounts for also `/lib`, `/usr/lib`, `/usr/src` and `/proc/swaps`

Maybe we should sync with that

internal/configmap/testdata/nfd.kubernetes.io_v1_nodefeaturediscovery.yaml
chr15p: Done. that field isn't actually being used from that test data, its just using the configData to populate the configmap, so I've changed it to future proof.

internal/controllers/nodefeaturediscovery_reconciler.go
chr15p: Done. Yes configmap *sigh*

internal/daemonset/worker.go
chr15p: Done

internal/daemonset/worker.go
chr15p: Done. If we're going to need them then it makes sense to add them in now rather then need another PR later.

internal/daemonset/daemonset.go
chr15p: I've left this as is for the moment.  I dont think we're intending to enable deprecated features, so I'm happy to remove it , unless @yevgeny-shnaidman disagrees?

yevgeny-shnaidman: @chr15p lets remove it

chr15p:
ping @marquiz
ping @yevgeny-shnaidman

PR #221, opened 2024-04-02
set permissions for the refactored controller

chr15p:
This will set up the various service accounts, roles and rolebindings needed by the new (refactored) controller.  These will all be applied at install (`make deploy`) time.
For clarity the current  manager setup (`config/rbac/role.yaml` and `role_binding.yaml`) are renamed, and each account is broken into its own set of files.

If applied before the new controller is made live the old controller will overwrite any changes here at reconcile time and continue to work (or at least will continue to be as broken as it is now...) so this PR can be applied safely without the refactor being completed.

### DIFF
--- a/config/crd/bases/nfd.k8s-sigs.io_nodefeatures.yaml
+++ b/config/crd/bases/nfd.k8s-sigs.io_nodefeatures.yaml
@@ -1,0 +1,111 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.9.2
+  creationTimestamp: null
+  name: nodefeatures.nfd.k8s-sigs.io
+spec:
+  group: nfd.k8s-sigs.io
+  names:
+    kind: NodeFeature
+    listKind: NodeFeatureList
+    plural: nodefeatures
+    singular: nodefeature
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: NodeFeature resource holds the features discovered for one node
+          in the cluster.
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: NodeFeatureSpec describes a NodeFeature object.
+            properties:
+              features:
+                description: Features is the full "raw" features data that has been
+                  discovered.
+                properties:
+                  attributes:
+                    additionalProperties:
+                      description: AttributeFeatureSet is a set of features having
+                        string value.
+                      properties:
+                        elements:
+                          additionalProperties:
+                            type: string
+                          type: object
+                      required:
+                      - elements
+                      type: object
+                    description: Attributes contains all the attribute-type features
+                      of the node.
+                    type: object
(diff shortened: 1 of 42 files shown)


config/rbac/gc_clusterrole.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: gc
```
marquiz, line 4: Should we use `nfd-` prefix for all the names. I think this is relevant especially for the cluster-scoped resources like this.

```suggestion
  name: nfd-gc
```

marquiz (commented):
/ok-to-test

config/rbac/gc_clusterrole.yaml
chr15p: when `make deploy` runs kustomize it appends `nfd-` to all the names in the config it generates  (set by the `namePrefix: nfd-` setting in config/default/kustomization.yaml), so all the roles, service-accoutns etc should be named nfd-*

marquiz: Hmm, ok. I still find it confusing that the clusterrole references a ServiceAccount with a different name than what's here 🤨 If the names are all replaced could we just use a placeholder (like `NAME`) for all of them to make it clear that it will be replaced. Or is it too obvious (to everyone else than me)?

yevgeny-shnaidman: maybe we might remove the namePrefix? IMHO the definitions of the roles/accounts/etc' become more readable, we don't need to take all the substitutions into account when reading the yamls..

marquiz: > IMHO the definitions of the roles/accounts/etc' become more readable, we don't need to take all the substitutions into account when reading the yamls..

I agree. Would make it more readable and easier to review and spot mistakes

chr15p: I agree, it is a bit confusing as is, while testing it took me a bit to figure out why I was getting names like `nfd-nfd-gc`   I left it in to try and keep the focus of the PR tight, however it sounds like we all agree that its worth doing, so I will make the change and update the PR.

marquiz: > I left it in to try and keep the focus of the PR tight,

Mm, that could also be done in a separate (prior) PR. If you like.

yevgeny-shnaidman:
@chr15p WDYT about creating sub-directories (master, worker, prune etc') under config/rbac ? will make it a little bit clearer to understand which roles/sa relevant for which component

chr15p:
I've moved the components into their own subdirectories of rbac, and the parts needed for the operator itself (manager and leader_election role/sa etc)  into a `core` directory so hopefully its a little more obvious what files are used be what.

I have also removed the `namePrefix:` and hardcoded all the names to start with nfd-  to be less confusing. Everything in my test setup is working and labels are being applied by the new controller and all the object names look good so I think I've caught all the name changes.

Lastly I've synced the clusterroles/roles with the helm charts.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-nfd/node-feature-discovery-operator
-  newTag: master
+  newName: registry.k8s.io/nfd/node-feature-discovery-operator
+  newTag: 0.4.2
```
marquiz, line 16: I'd suggest to leave this as it was

marquiz (commented):
Wow @chr15p now this is so much easier for me to read 🚀 One small nit below but otherwise looks good

config/manager/kustomization.yaml
chr15p: good catch. This gets set by the `make deploy` process so they keep creeping in while I'm testing, I obviously missed this one and committed it by mistake. I have reverted it.

ArangoGutierrez (approved):
Approved, just a nit, could you quash the commits? after that I'll lgtm

chr15p:
Squished the commits and reworded the commit message to cover all the changes.

ArangoGutierrez (approved):
I'll leave the lgtm to @marquiz

marquiz (approved):
Good job on this @chr15p. A bit of extra work but I think it was worth it 😇
/lgtm

PR #224, opened 2024-04-14
Adding prune job to the finalization flow

yevgeny-shnaidman:
1) in case PruneOnDelete defined, prune job is created during
   finalization flow.
2) the finalizer is not deleted from NFD object until prune job is
   completed successfully
3) changing the filtering to responds to events only in case of NFD
   owned objects


yevgeny-shnaidman:
ping @ArangoGutierrez

yevgeny-shnaidman:
ping @marquiz

marquiz (approved):
Thanks @yevgeny-shnaidman, looks good

@ArangoGutierrez PTAL

ArangoGutierrez (approved):
/lgtm

PR #225, opened 2024-04-16
adding golangci-lint to the skipper Dockerfile

yevgeny-shnaidman:


Dockerfile.node-feature-discovery-operator-build
```diff
@@ -10,5 +10,6 @@ RUN ["apk", "add", "bash", "make", "docker", "curl", "shadow", "git"]
 RUN go install go.uber.org/mock/mockgen@v0.3.0
 RUN curl -L "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" -o /tmp/kubectl
 RUN install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl
+RUN go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.56.1
```
ArangoGutierrez, line 13: any reason why not the latest? `1.57.2`

ArangoGutierrez (changes requested):

Dockerfile.node-feature-discovery-operator-build
yevgeny-shnaidman: Just to be aligned with the one in the Makefile.

ArangoGutierrez: I won't be against an extra commit in this PR bumping that version

yevgeny-shnaidman: Fixed

ArangoGutierrez (approved):
I'll leave the lgtm to @marquiz

marquiz (approved):
/lgtm

PR #227, opened 2024-05-07
Adding NODE_ADDRESS env variable to topology DS to support default

yevgeny-shnaidman:
configz

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -75,7 +75,7 @@ func (d *daemonset) SetTopologyDaemonsetAsDesired(ctx context.Context, nfdInstan
 							"nfd-topology-updater",
 						},
 						Args:            getArgs(nfdInstance),
-						Env:             getEnvs(),
+						Env:             getTopologyEnvs(),
 						SecurityContext: getSecurityContext(),
 						VolumeMounts:    getVolumeMounts(),
 					},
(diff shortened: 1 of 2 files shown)


internal/daemonset/daemonset.go
```diff
@@ -128,6 +128,27 @@ func getEnvs() []corev1.EnvVar {
 	}
 }
 
+func getTopologyEnvs() []corev1.EnvVar {
```
ArangoGutierrez, line 131: you could reuse `getEnv` and simply `append` `NODE_ADDRESS` or 
```go
type NodeEnv struct {
	Name      string
	FieldPath string
}

func getWorkerEnvs(envs ...NodeEnv) []corev1.EnvVar {
	envList := []corev1.EnvVar{}

	for _, env := range envs {
		envList = append(envList, corev1.EnvVar{
			Name: env.Name,
			ValueFrom: &corev1.EnvVarSource{
				FieldRef: &corev1.ObjectFieldSelector{
					FieldPath: env.FieldPath,
				},
			},
		})
	}

	return envList
}
```

ArangoGutierrez (changes requested):

internal/daemonset/daemonset.go
yevgeny-shnaidman: fixed, used the first option , seemed a little bit more readable

ArangoGutierrez (approved):
/lgtm

PR #228, opened 2024-05-07
adding permissions to the NFD controller

yevgeny-shnaidman:
adding kubebuiler options for creating RBAC permissions needed for NFD controller, and creating the actula role yaml

### DIFF
--- a/internal/controllers/nodefeaturediscovery_reconciler.go
+++ b/internal/controllers/nodefeaturediscovery_reconciler.go
@@ -51,9 +51,9 @@ type nodeFeatureDiscoveryReconciler struct {
 }
 
 func NewNodeFeatureDiscoveryReconciler(client client.Client, deploymentAPI deployment.DeploymentAPI, daemonsetAPI daemonset.DaemonsetAPI,
-	configmapAPI configmap.ConfigMapAPI, jobAPI job.JobAPI, scheme *runtime.Scheme) nodeFeatureDiscoveryReconciler {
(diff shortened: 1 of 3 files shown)


internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -99,7 +99,11 @@ func isControlledByNFD(obj client.Object) bool {
 // +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
```
ArangoGutierrez, line 102: Batch?

ArangoGutierrez (changes requested):
Why do we need RBAC for `batch`?

internal/controllers/nodefeaturediscovery_reconciler.go
yevgeny-shnaidman: we are handing prune Jobs

ArangoGutierrez: ah the prune stuff..... my bad

ArangoGutierrez (approved):
/lgtm

PR #229, opened 2024-05-08
updating skipper configuration

yevgeny-shnaidman:
1) correcting the env variable needed for image creation
2) adding configuration to allow deploying NFD on minikube


ArangoGutierrez (approved):
/lgtm

PR #230, opened 2024-05-13
Fixing prune job implementation

yevgeny-shnaidman:
No need to explicitly delete the prune job, since it will be deleted by the K8S scheduler once the NFD CR is removed from etcd. Additionally, if the job is explicitly deleted, then the code will attempt to recreate it a second time while the NFD CR is being deleted.

### DIFF
--- a/internal/controllers/nodefeaturediscovery_reconciler.go
+++ b/internal/controllers/nodefeaturediscovery_reconciler.go
@@ -330,23 +330,17 @@ func (nfdh *nodeFeatureDiscoveryHelper) handlePrune(ctx context.Context, nfdInst
 		return false, fmt.Errorf("failed to get nfd-prune job: %w", err)
 	}
 
-	deleteJob := false
 	var returnErr error
 	done := false
 	if pruneJob.Status.Succeeded > 0 {
-		deleteJob = true
 		done = true
 	}
 	if pruneJob.Status.Failed > 0 {
-		deleteJob = true
 		returnErr = fmt.Errorf("prune job's pod has failed")
 	}
-	if deleteJob {
-		err = nfdh.jobAPI.DeleteJob(ctx, pruneJob)
-		if err != nil {
-			return false, fmt.Errorf("failed to delete nfd-prune job: %w", err)
-		}
-	}
+
+	// no need to explicitly delete Prune job,
+	// it will be deleted by K8S scheduler once NFD CR is deleted from etcd
 	return done, returnErr
 }
 
--- a/internal/controllers/nodefeaturediscovery_reconciler_test.go
+++ b/internal/controllers/nodefeaturediscovery_reconciler_test.go
@@ -758,7 +758,7 @@ var _ = Describe("handlePrune", func() {
 		Expect(done).To(BeFalse())
 	})
 
-	DescribeTable("prune job exsists flows", func(podFailed, podSucceeded, deleteFailed bool) {
+	DescribeTable("prune job exsists flows", func(podFailed, podSucceeded bool) {
 		nfdCR.Spec.PruneOnDelete = true
 		foundJob := batchv1.Job{}
 		if podFailed {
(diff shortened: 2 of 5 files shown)


internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -330,23 +330,25 @@ func (nfdh *nodeFeatureDiscoveryHelper) handlePrune(ctx context.Context, nfdInst
 		return false, fmt.Errorf("failed to get nfd-prune job: %w", err)
 	}
 
-	deleteJob := false
+	//deleteJob := false
```
ArangoGutierrez, line 333: Why leave this line commented and not simply remove it?

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -330,23 +330,25 @@ func (nfdh *nodeFeatureDiscoveryHelper) handlePrune(ctx context.Context, nfdInst
...
 	var returnErr error
 	done := false
 	if pruneJob.Status.Succeeded > 0 {
-		deleteJob = true
+		//deleteJob = true
```
ArangoGutierrez, line 337: Why leave this line commented and not simply remove it?

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -330,23 +330,25 @@ func (nfdh *nodeFeatureDiscoveryHelper) handlePrune(ctx context.Context, nfdInst
...
 		done = true
 	}
 	if pruneJob.Status.Failed > 0 {
-		deleteJob = true
+		//deleteJob = true
```
ArangoGutierrez, line 341: Why leave this line commented and not simply remove it?

internal/controllers/nodefeaturediscovery_reconciler.go
```diff
@@ -330,23 +330,25 @@ func (nfdh *nodeFeatureDiscoveryHelper) handlePrune(ctx context.Context, nfdInst
...
 		returnErr = fmt.Errorf("prune job's pod has failed")
 	}
-	if deleteJob {
-		err = nfdh.jobAPI.DeleteJob(ctx, pruneJob)
-		if err != nil {
-			return false, fmt.Errorf("failed to delete nfd-prune job: %w", err)
+	/*
```
ArangoGutierrez, line 344: Why leave this line commented and not simply remove it?

ArangoGutierrez (changes requested):

internal/controllers/nodefeaturediscovery_reconciler.go
yevgeny-shnaidman: my bad, forgot to remove it, fixed

internal/controllers/nodefeaturediscovery_reconciler.go
yevgeny-shnaidman: my bad, forgot to remove it, fixed

internal/controllers/nodefeaturediscovery_reconciler.go
yevgeny-shnaidman: my bad, forgot to remove it, fixed

internal/controllers/nodefeaturediscovery_reconciler.go
yevgeny-shnaidman: my bad, forgot to remove it, fixed

ArangoGutierrez (changes requested):
Please squash the 2 commits and fix the commit/PR message:
"No need to explicitly delete the prune job, since it will be deleted by the K8S scheduler once the NFD CR is removed from etcd. Additionally, if the job is explicitly deleted, then the code will attempt to recreate it a second time while the NFD CR is being deleted."

yevgeny-shnaidman:
> Please squash the 2 commits and fix the commit/PR message: "No need to explicitly delete the prune job, since it will be deleted by the K8S scheduler once the NFD CR is removed from etcd. Additionally, if the job is explicitly deleted, then the code will attempt to recreate it a second time while the NFD CR is being deleted."

Done

ArangoGutierrez (approved):
/lgtm

PR #232, opened 2024-05-15
Pivoting to new operator implementation

yevgeny-shnaidman:
Starting from this commit, operator will use the new controller implementation. The old implementation code will be deleted in the subsequent PRs

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,7 +22,7 @@ import (
 	"os"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/klog/v2"
+	"k8s.io/klog/v2/textlogger"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


marquiz:
Thanks @yevgeny-shnaidman for the PR. I'll check it in detail a bit later. First question (after a 30s skim-through): are the logger changes related/dependent on this (or could they be a separate PR)?

yevgeny-shnaidman:
> Thanks @yevgeny-shnaidman for the PR. I'll check it in detail a bit later. First question (after a 30s skim-through): are the logger changes related/dependent on this (or could they be a separate PR)?

They are dependent on switching to the new controller. Unless we do them, we will see some warnings in the manager log

marquiz:
> They are dependent on switching to the new controller. Unless we do them, we will see some warnings in the manager log

Kinda useless nitting, but where do we get warnings? Looking at the PR now we only change logger in main.go which shouldn't affect any other parts of the codebase 🧐 So it looks to me like the logger changes could/should be just a separate PR.

yevgeny-shnaidman:
> > They are dependent on switching to the new controller. Unless we do them, we will see some warnings in the manager log
> 
> Kinda useless nitting, but where do we get warnings? Looking at the PR now we only change logger in main.go which shouldn't affect any other parts of the codebase 🧐 So it looks to me like the logger changes could/should be just a separate PR.

We get the warning in the manager log, but those warnings are coming from the controller-runtime code. I can split this PR into 2 ( first the actual pivot and then the logging PR), but i wanted to keep it altogether so that after this PR the operator can be seen deployed and running with logs.

marquiz (approved):
Thanks @yevgeny-shnaidman. I'm good to merge this as is 

/assign @ArangoGutierrez

yevgeny-shnaidman:
ping @ArangoGutierrez

ArangoGutierrez:
> ping @ArangoGutierrez

Sorry I was traveling last week, will review this one this week

yevgeny-shnaidman:
ping @ArangoGutierrez

PR #233, opened 2024-05-23
Adding Getters to the Deployment and Daemonset interfaces

yevgeny-shnaidman:
Getter functions are a useful interface, and will be used in the status implementation


marquiz (approved):
/lgtm

PR #234, opened 2024-05-30
Adding NFD CR status handling

yevgeny-shnaidman:
Status conditions are going to reflect the status
of the master/worker/gc/topology deployments/daemonsets. No need to reflect the status of SA/Roles/ etc', since they are going to be created during operator deployment


yevgeny-shnaidman:
ping @marquiz 
ping @ArangoGutierrez

yevgeny-shnaidman:
ping @marquiz
ping @ArangoGutierrez

internal/status/status.go
```diff
@@ -0,0 +1,318 @@
...
+	if ds.Status.NumberReady == ds.Status.DesiredNumberScheduled {
+		return conditionStatusAvailable, ""
+	}
+	return conditionStatusProgressing, "ds is progressing"
+}
+
+func getDeploymentConditions(dep *appsv1.Deployment) (string, string) {
+	if dep.Status.AvailableReplicas == 0 {
+		return conditionStatusDegraded, "number of available pods is 0"
+	}
```
marquiz, line 215: Do we have support for more than one replica (in the operator)? Should we  add similar StatusProgressing condition as for the daemonset?

internal/status/status.go
```diff
@@ -0,0 +1,318 @@
...
+	if conditionsStatus == conditionStatusDegraded {
+		return getDegradedConditions(dsDegradedReason, message)
+	} else if conditionsStatus == conditionStatusProgressing {
+		return getProgressingConditions(dsProgressingReason, message)
+	}
+	return nil
+}
+
+func (sh *statusHelper) getDeploymentNotAvailableConditions(ctx context.Context, deploymentNamespace, deploymentName string) []metav1.Condition {
+	failedToGetDeploymentReason := conditionFailedGettingNFDMasterDeployment
```
marquiz, line 178: We use this for GC as well so we shouldn't unconditionally do `conditionFailedGettingNFDMasterDeployment`?

internal/status/status.go
```diff
@@ -0,0 +1,318 @@
...
+	conditionFailedGettingNFDWorkerDaemonSet = "FailedGettingNFDWorkerDaemonSet"
+	conditionNFDWorkerDaemonSetDegraded      = "NFDWorkerDaemonSetDegraded"
+	conditionNFDWorkerDaemonSetProgressing   = "NFDWorkerDaemonSetProgressing"
+
+	conditionFailedGettingNFDTopologyDaemonSet = "FailedGettingNFDTopologyDaemonSet"
+	conditionNFDTopologyDaemonSetDegraded      = "NFDTopologyDaemonSetDegraded"
+	conditionNFDTopologyDaemonSetProgressing   = "NFDTopologyDaemonSetProgressing"
+
+	conditionFailedGettingNFDMasterDeployment  = "FailedGettingNFDMasterDeployment"
+	conditionNFDWorkerMasterDeploymentDegraded = "NFDMasterDeploymentDegraded"
```
marquiz, line 46: WorkerMaster?

internal/status/status.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func newStatusHelperAPI(deploymentAPI deployment.DeploymentAPI, daemonsetAPI daemonset.DaemonsetAPI) statusHelperAPI {
+	return &statusHelper{
+		deploymentAPI: deploymentAPI,
+		daemonsetAPI:  daemonsetAPI,
+	}
+}
+
+func (sh *statusHelper) getDaemonSetNotAvailableConditions(ctx context.Context, dsNamespace, dsName string) []metav1.Condition {
```
marquiz, line 155: nit: Maybe rename this to indicate that this is solely for the nfd-worker daemonset. Or alternatively make the function more generic?

```suggestion
func (sh *statusHelper) getWorkerDaemonSetNotAvailableConditions(ctx context.Context, dsNamespace, dsName string) []metav1.Condition {
```

marquiz (commented):
Thanks @yevgeny-shnaidman for the PR. Looks good in general. Just a few comments/questions below

internal/controllers/nodefeaturediscovery_reconciler_test.go
```diff
@@ -788,3 +789,73 @@ var _ = Describe("handlePrune", func() {
...
+		gomock.InOrder(
+			mockStatus.EXPECT().GetConditions(ctx, &nfdCR).Return(newConditions),
+			mockStatus.EXPECT().AreConditionsEqual(newConditions, nfdCR.Status.Conditions).Return(true),
+		)
+
+		err := nfdh.handleStatus(ctx, &nfdCR)
+		Expect(err).To(BeNil())
+	})
+
+	It("conditions are equal, status update is needed", func() {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: yes, you are correct.Fixed

internal/controllers/nodefeaturediscovery_reconciler_test.go
```diff
@@ -788,3 +789,73 @@ var _ = Describe("handlePrune", func() {
...
+			mockStatus.EXPECT().AreConditionsEqual(newConditions, nfdCR.Status.Conditions).Return(false),
+			clnt.EXPECT().Status().Return(statusWriter),
+			statusWriter.EXPECT().Patch(ctx, &expectedNFD, gomock.Any()).Return(nil),
+		)
+
+		err := nfdh.handleStatus(ctx, &nfdCR)
+		Expect(err).To(BeNil())
+	})
+
+	It("conditions are equal, status update failed", func() {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

internal/status/status.go
yevgeny-shnaidman: Currently operator deploys with replica of 1 and it is not configurable by user.  
In regards to Progressing vs Degraded: the distinction is blurred, but in this PR i was adhering to the current implementation. It is not urgent, but i think that we can discuss if we really need both Degraded and Progressing.How we distinguish between the 2: in both cases we don't have enough pods running, and K8S will try to bring up additional pods, so we are both in Degraded state and in Progressing state...

internal/status/status.go
yevgeny-shnaidman: The function checks what is the deployment (master or gc) and if needed changes the Reason. Please see line 181

yevgeny-shnaidman: actually, my current code is sorta of ugly, i am changing it

internal/status/status.go
yevgeny-shnaidman: Thanks, fixed.

internal/status/status.go
yevgeny-shnaidman: This function is generic (it deals with both worker and topolgy daemonsets), but i will re-write it to make the code clearer

yevgeny-shnaidman:
@marquiz, pushed my changes , please review

marquiz (approved):
Thanks @yevgeny-shnaidman for the update. My previous review was a bit hasty but the code is much clearer (to me at least) now.

/lgtm

PR #235, opened 2024-05-31
sync nodefeaturerules crd config with helm

chr15p:
The nodefeaturerules crd differs between _config/_ and the _deploy/helm/_  directorys, with the taints structure being missing. This PR brings the _config/_ version into sync with the _helm/_ version by adding the missing fields.

(I have checked the other CRDs and they seem to be ok, its just rules that is out of sync)


chr15p:
ping @marquiz
ping @yevgeny-shnaidman

marquiz (approved):
Thanks @chr15p, looks fine to me
/assign @yevgeny-shnaidman

PR #236, opened 2024-06-05
sync operator nfr crds with operand helm nfr crds

chr15p:
sync operator nfr crds with operand helm nfr crds  and also standardise on nodefeaturerules.nfd.k8s-sigs.io

The crds here do not support `annotations:`, `extendedResources:` and `matchFeatures.matchName` in their `rules:`.   They are supported by the CRDs supplied by the operand helm charts (which as I understand it are correct) so it make sense to sync them over. This PR does that.

We also have confusion on the api version for rules with  `nfd.k8s-sigs.io/v1alpha1`  being used in most places, and `nfd.kubernetes.io/v1alpha1` in the samples, and manifest files.  k8s-sigs.io is used by the operand helm charts so this PR updates to that, removing the extra crd file (that wasn't referenced by kustomize anyway), and renaming the sample yaml file.


marquiz (approved):
Thanks @chr15p for the PR. I didn't compare the CRD definitions but looks good 😅
/assign @yevgeny-shnaidman

PR #237, opened 2024-06-05
add enableTaints option to enable taint support in the crd

chr15p:
It is currently not possible to enable taints support from the operator, this PR adds that behaviour by adding an `spec.enableTaints:` boolean option to the nodefeaturediscoveries crd. This option is false by default, but when set to true adds the `--enable-taints` switch to the master args.

This is false by default, and commented out in the sample config because it is still rated as "experimental" but this allows brave users to test it.


marquiz (approved):
/lgtm

PR #238, opened 2024-06-06
Getting rid of pkg/utils

yevgeny-shnaidman:
pkg utils is used for solely for getting the WATCH_NAMESPACE env variable . We don't need a separate public package just for that functionality. Moving the function into the main. In addition, since WATCH_NAMESPACEmust be set in the deployment, the operator will fail in case the variable is not defined, instead of just logging a warrning


marquiz (commented):
Should we drop `pkg/utils/utils.go` file?

yevgeny-shnaidman:
> Should we drop `pkg/utils/utils.go` file?

We will do it in the same PR for all the irrelevant directories (pkg, controllers, build/assets)

marquiz (approved):
/lgtm

PR #240, opened 2024-06-10
Getting rid of pkg/version

yevgeny-shnaidman:
package version is used solely to provide the code version of the operator code when executed with version argument. No need to have a public package just for that code, implementation is moved into the main package


marquiz (approved):
/lgtm

PR #241, opened 2024-06-13
Removing old/unused packages

yevgeny-shnaidman:
The following packages are removed since they are no longer in user with the new implementation
1) controllers
2) build/assets
3) pkg


yevgeny-shnaidman:
ping @marquiz 
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #242, opened 2024-06-13
add nodefeaturegroups support

chr15p:
nodefeaturegroups have been added to the operand, so this PR adds the crd and the required permissions in the operator install to prevent crashing


chr15p:
@ArangoGutierrez or @marquiz could you give this an `/approve` please (assuming you do obviously!)

PR #243, opened 2024-06-14
sync helm chart crds/rbac with the kustomize files

chr15p:
The helm charts are out of sync with those used by the kustomize installer, this commit copies them over and converts them to use helm variables.  This should now mean you get the same set of crds, serviceaccounts, bindings, etc. whichever of the two installation methods you use, and nfd should Just Work (tm)  either way.

Hopefully by breaking the objects out into the multi-file layout used by kustomize it will be easier to spot differences and keep the two installation methods in sync in the future.

(note: Most rbac objects should have the chart fullname in their name but the serviceaccount names are hardwired in the operator reconciliation code so they can't be templated and are left as nfd-worker, nfd-master etc)


chr15p:
@ArangoGutierrez or @marquiz  this one also needs an `/approve` please (again, assuming you do!)

ArangoGutierrez (approved):

PR #255, opened 2025-03-02
Adding Tolerations fields to the NFD CRD

yevgeny-shnaidman:
Currently, while deploying operand only, user can specify tolerations for both master depoyment and worker daemonsets. We want to add this functionlity to operator deployment also, using dedicated fields in the CRD


ArangoGutierrez (approved):
/lgtm

PR #256, opened 2025-03-03
Adding tolerations implementation to master and worker pod

yevgeny-shnaidman:
This commit add tolerations defined in the WorkerTolerations field of the NFD CR to the worker daemonset tolerations definition.It also does the same for master deployment based on the MasterTolerations field of the NFD CR

### DIFF
--- a/internal/daemonset/worker.go
+++ b/internal/daemonset/worker.go
@@ -20,6 +20,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	"k8s.io/utils/ptr"
+	nfdv1 "sigs.k8s.io/node-feature-discovery-operator/api/v1"
 )
 
 func getWorkerAffinity() *corev1.Affinity {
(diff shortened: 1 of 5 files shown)


marquiz:
I'll be OoO for a week. Looking at @arangogutierrez if you have time for this :blush:

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-nfd/node-feature-discovery-operator
-  newTag: master
+  newName: registry.k8s.io/nfd/node-feature-discovery-operator
+  newTag: 0.4.2
```
ArangoGutierrez, line 16: I think this should not be changed for master

ArangoGutierrez (changes requested):

internal/daemonset/worker.go
```diff
@@ -212,3 +213,14 @@ func getWorkerVolumes() []corev1.Volume {
 func getWorkerLabelsAForApp(name string) map[string]string {
 	return map[string]string{"app": name}
 }
+
+func getWorkerToleration(nfdInstance *nfdv1.NodeFeatureDiscovery) []corev1.Toleration {
```
ArangoGutierrez, line 217: I agree with the Copilot, let's rename this func

ArangoGutierrez (changes requested):

internal/daemonset/worker.go
yevgeny-shnaidman: Fixed

config/manager/kustomization.yaml
yevgeny-shnaidman: Fixed

ArangoGutierrez (approved):
/lgtm

PR #261, opened 2025-03-23
Configuring the worker DS to target only "linux" nodes

yevgeny-shnaidman:


internal/daemonset/testdata/test_worker_daemonset.yaml
```diff
@@ -26,9 +26,17 @@ spec:
             - matchExpressions:
               - key: node-role.kubernetes.io/master
                 operator: DoesNotExist
+              - key: beta.kubernetes.io/os
```
marquiz, line 29: Shouldn't we use the GA label?

```suggestion
              - key: kubernetes.io/os
```

yevgeny-shnaidman: yes, we should. Fixed it

marquiz (approved):
Thank you @yevgeny-shnaidman. Looks good now
/lgtm

PR #265, opened 2025-08-04
Adding liveliness readiness probes to the GC

yevgeny-shnaidman:


yevgeny-shnaidman:
> Hmm, in nfd (the next release) we've moved to http and /healthz in health probes. Probably the operator dev branch should reflect that.

Thanks! changed it to http, please take a look

yevgeny-shnaidman:
@marquiz please take a look again

PR #267, opened 2026-03-03
Allowing to configure priority class for the worker pods

yevgeny-shnaidman:
This PR adds a WorkerPriorityClassName field to the operand configuration in NFD, to allow user to configure priority class name for the worker pods


marquiz (approved):
Thanks @yevgeny-shnaidman, looks good to me

/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

